PR #36, opened 2023-04-21
[Task] Removed deprecated `attributes` field in link editable

Corepex:


Corepex:
removed translations in https://github.com/pimcore/pimcore/pull/15038/commits/10db8a23581801eaa630c1ef698b152d5dc6e2a6

PR #37, opened 2023-04-21
[Task]: Remove system/admin settings remnants

### LINKED ISSUE #14997: [Task]: Remove system/admin settings remnants
### Improvement description

Remove system.yaml:
https://github.com/pimcore/pimcore/blob/a6cc3a2b37cc415d74f4f1d14e8c026d5c1cd382/bundles/InstallBundle/src/Command/InstallCommand.php#L196

https://github.com/pimcore/pimcore/blob/03a75921d31a7307b9dfbcb75f3c8633f0247a95/lib/Config/LocationAwareConfigRepository.php#L279

Check `getSystemConfiguration` calls for admin bundle configurations. Then remove:
https://github.com/pimcore/pimcore/blob/85d6471de3a187cc32db282dc4be2179a4da50a1/lib/Config.php#L155

lukmzig:
Partially resolves https://github.com/pimcore/pimcore/issues/14997

### DIFF
--- a/src/Controller/Admin/UserController.php
+++ b/src/Controller/Admin/UserController.php
@@ -1120,9 +1120,7 @@ class UserController extends AdminController implements KernelControllerEventInt
      */
     public function getDefaultKeyBindingsAction(Request $request): JsonResponse
     {
-        $data = UserHelper::getDefaultKeyBindings();
-
-        return $this->adminJson(['success' => true, 'data' => $data]);
+        return $this->adminJson(['success' => true, 'data' => UserHelper::getDefaultKeyBindings()]);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Controller/Admin/UserController.php
```diff
@@ -637,7 +638,7 @@ public function getCurrentUserAction(Request $request): Response
         $userData = $user->getObjectVars();
         $contentLanguages = Tool\Admin::reorderWebsiteLanguages($user, Tool::getValidLanguages());
         $userData['contentLanguages'] = $contentLanguages;
-        $userData['keyBindings'] = $user->getKeyBindings();
+        $userData['keyBindings'] = $user->getKeyBindings() ?? UserHelper::getDefaultKeyBindings();
```
dvesh3: I would implement the fallback in the helper by passing a user object.

mattamon (changes requested):
Do we need all parts of the helper.
The default keybindings should be defined in the config itself -> https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/config/pimcore/default.yaml#L22 So we could remove the legacy ones.

And the strict method is also not needed anymore since all values have now the default false unless set to true.
Check the Configuration.php -> https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/src/DependencyInjection/Configuration.php#L463

PR #29, opened 2023-04-19
Decouple from the core

dvesh3:
Related to https://github.com/pimcore/pimcore/pull/14918 & https://github.com/pimcore/pimcore/pull/14971

PR #39, opened 2023-04-21
[Github Action] Setup codeception and add test from core. 

### LINKED ISSUE #3: [Github Action] Setup codeception

martineiber:
Resolves #3 

-----

Add test form core bundle. 
Waiting for https://github.com/pimcore/pimcore/pull/14971


.github/ci/files/config/services.yaml
```diff
@@ -0,0 +1,87 @@
...
+        autoconfigure: true
+
+    #
+    # Notification Services
+    #
+
+    Pimcore\Model\Notification\Service\NotificationService:
+        public: true
+
+    Pimcore\Model\Notification\Service\UserService: ~
```
dvesh3: these can go. see https://github.com/pimcore/admin-ui-classic-bundle/pull/29/files#diff-c47b03734cd2b4337b345eeba955637ba790d51fd98979f6d366d4acbdb104e0L10

.github/workflows/codeception.yml
```diff
@@ -0,0 +1,78 @@
...
+    codeception-tests:
+        name: "Codeception tests"
+        runs-on: "ubuntu-20.04"
+        continue-on-error: ${{ matrix.experimental }}
+        env:
+            PIMCORE_TEST_DB_DSN: "mysql://root@127.0.0.1:33006/pimcore_test?serverVersion=${{ matrix.server_version }}"
+        strategy:
+            matrix:
+                include:
+                    - { php-version: 8.1, database: "mariadb:10.7", server_version: "10.7.7-MariaDB-1:10.7.7+maria~ubu2004", dependencies: highest, experimental: false }
```
dvesh3:
```suggestion
                    - { php-version: 8.1, database: "mariadb:10.7", server_version: "10.7.7-MariaDB-1:10.7.7+maria~ubu2004", dependencies: lowest, experimental: false }
```
Run at-least one against lowest

composer.json
```diff
@@ -9,15 +9,17 @@
       "*": "dist"
     }
   },
+  "prefer-stable": true,
+  "minimum-stability": "dev",
   "require": {
-    "pimcore/pimcore": "^11.0",
+    "pimcore/pimcore": "^11.x-dev",
```
dvesh3:
```suggestion
    "pimcore/pimcore": "^11.0",
```
Any specific reason for requiring x-dev?

martineiber: No, this was just for some testing. Thx for the catch.

.github/ci/files/config/services.yaml
martineiber: Why was it removed. it is still used here `src/Controller/Admin/NotificationController.php` right?

dvesh3: because the service class is in the core and admin bundle can still use the service from core.

dvesh3 (approved):

PR #45, opened 2023-04-25
[Security] Fixed xss in translations panel

Corepex:


Corepex:
belongs to: https://github.com/pimcore/pimcore/pull/15038

PR #54, opened 2023-04-26
Fix targeting page preview and move bundle specific css

aryaantony92:
Related to https://github.com/pimcore/personalization-bundle/pull/30


mattamon (approved):
LGTM

PR #57, opened 2023-04-28
[Security] Set firewall settings parameter through extension

dvesh3:


kingjia90 (approved):
LGTM so far, merging when https://github.com/pimcore/openid-connect/pull/73 is also ready

PR #59, opened 2023-05-03
Remove deprecated code and make scriptProxyAction() as internal

### LINKED ISSUE #16: Remove deprecated code from scrript proxy action
See: 
https://github.com/pimcore/admin-ui-classic-bundle/blob/d8bb3749655936727e448ae4c8bd4a81ba780bf2/src/Controller/Admin/MiscController.php#L149

We should also check whether we still need the script-proxy anyway ... ideally we could get rid of it.

aryaantony92:
Resolves #16 
Please see https://github.com/pimcore/pimcore/pull/15093 for the upgrade note


mattamon (approved):
LGTM

PR #63, opened 2023-05-08
[Bug] Removed on `change` sanitizer

Corepex:
Followup to: https://github.com/pimcore/pimcore/pull/15111


Corepex:
obsoleted by: https://github.com/pimcore/admin-ui-classic-bundle/pull/77

PR #66, opened 2023-05-09
Remove personalization editable code

### LINKED ISSUE #34: [Bug] Fix the deletion of editable Data
Steps to reproduce

1. Open a document
2. Select a target group variant by clicking 'Edit content for Target Group' menu in the bottom
3. Edit the content of an editable in the doument and save the document
4. Now, reload the default variant of the document again

The data of the editable which we edited in the default variant of the document will be lost.

aryaantony92:
Resolves https://github.com/pimcore/personalization-bundle/issues/34

### DIFF
--- a/src/Controller/Admin/Document/DocumentControllerBase.php
+++ b/src/Controller/Admin/Document/DocumentControllerBase.php
@@ -21,6 +21,7 @@ use Pimcore\Bundle\AdminBundle\Controller\Traits\DocumentTreeConfigTrait;
 use Pimcore\Bundle\AdminBundle\Controller\Traits\UserNameTrait;
 use Pimcore\Bundle\AdminBundle\Event\AdminEvents;
 use Pimcore\Bundle\AdminBundle\Event\ElementAdminStyleEvent;
+use Pimcore\Bundle\PersonalizationBundle\Model\Document\Targeting\TargetingDocumentInterface;
 use Pimcore\Controller\KernelControllerEventInterface;
 use Pimcore\Controller\Traits\ElementEditLockHelperTrait;
 use Pimcore\Logger;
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/Document/DocumentControllerBase.php
```diff
@@ -170,16 +170,6 @@ protected function addSettingsToDocument(Request $request, Model\Document $docum
...
     {
         if ($document instanceof Model\Document\PageSnippet) {
-            // if a target group variant get's saved, we have to load all other editables first, otherwise they will get deleted
-
-            if ($request->get('appendEditables')) {
-                // ensure editable are loaded
-                $document->getEditables();
-            } else {
-                // ensure no editables (e.g. from session, version, ...) are still referenced
-                $document->setEditables(null);
```
dvesh3, line 180: It resurfaces this [bug ](https://github.com/pimcore/pimcore/issues/9749)again. we should at-least reset the editables when no personalization bundle is installed.

aryaantony92: Thanks for pointing out. Will adapt the change.

src/Controller/Admin/Document/DocumentControllerBase.php
```diff
@@ -170,16 +171,13 @@ protected function addSettingsToDocument(Request $request, Model\Document $docum
     protected function addDataToDocument(Request $request, Model\Document $document): void
     {
         if ($document instanceof Model\Document\PageSnippet) {
-            // if a target group variant get's saved, we have to load all other editables first, otherwise they will get deleted
-
-            if ($request->get('appendEditables')) {
-                // ensure editable are loaded
+            if($request->get('appendEditables') || (interface_exists(TargetingDocumentInterface::class) && $document instanceof TargetingDocumentInterface)) {
```
dvesh3, line 174: Is if part really required? since the target document requested is served through personalization bundle TargetingPageController or we just keep the else part here?

aryaantony92: If we skip the section $request->get('appendEditables') in the if part, then the content/document variants for each target group would not be saved

PR #69, opened 2023-05-10
[Events] Trigger Login Redirect event on login page not on authentication

dvesh3:
Related to https://github.com/pimcore/pimcore/pull/15137

### DIFF
--- a/src/Controller/Admin/LoginController.php
+++ b/src/Controller/Admin/LoginController.php
@@ -61,7 +61,7 @@ class LoginController extends AdminAbstractController implements KernelControlle
         protected ResponseHelper $responseHelper,
         protected TranslatorInterface $translator,
         protected PimcoreBundleManager $bundleManager,
-        protected EventDispatcherInterface $eventDispatcher,
+        protected EventDispatcherInterface $eventDispatcher
     ) {
     }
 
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/LoginController.php
```diff
@@ -181,21 +187,14 @@ public function logoutAction(): void
      */
     public function loginCheckAction(Request $request, EventDispatcherInterface $dispatcher): RedirectResponse
```
kingjia90:
```suggestion
    public function loginCheckAction(Request $request): RedirectResponse
```

src/Controller/Admin/LoginController.php
```diff
@@ -60,7 +60,8 @@ class LoginController extends AdminAbstractController implements KernelControlle
     public function __construct(
         protected ResponseHelper $responseHelper,
         protected TranslatorInterface $translator,
-        protected PimcoreBundleManager $bundleManager
+        protected PimcoreBundleManager $bundleManager,
+        protected EventDispatcherInterface $eventDispatcher,
```
kingjia90:
```suggestion
        protected EventDispatcherInterface $eventDispatcher
```

src/Controller/Admin/LoginController.php
kingjia90: but should we remove this dummy route then?

src/Controller/Admin/LoginController.php
```diff
@@ -181,21 +187,14 @@ public function logoutAction(): void
...
-        $event = new LoginRedirectEvent('pimcore_admin_login', [
-            'perspective' => strip_tags($request->get('perspective', '')),
-        ]);
-        $dispatcher->dispatch($event, AdminEvents::LOGIN_REDIRECT);
-
-        $url = $this->generateUrl($event->getRouteName(), $event->getRouteParams());
-
         // just in case the authenticator didn't redirect
-        return new RedirectResponse($url);
+        return new RedirectResponse($this->generateUrl('pimcore_admin_login'));
```
kingjia90:
```suggestion
        return new RedirectResponse($this->generateUrl('pimcore_admin_login', ['perspective' => strip_tags($request->get('perspective', ''))]));
```
this is needed, right? otherwise the perspective would be "lost"

kingjia90 (changes requested):

PR #72, opened 2023-05-11
[Task] Follow up permission restructure

mattamon:
Follow up to https://github.com/pimcore/admin-ui-classic-bundle/pull/55

### DIFF
--- a/public/js/pimcore/settings/user/role/settings.js
+++ b/public/js/pimcore/settings/user/role/settings.js
@@ -73,9 +73,6 @@ pimcore.settings.user.role.settings = Class.create({
         var sectionArray = [];
         for (var i = 0; i < this.data.availablePermissions.length; i++) {
             let section = this.data.availablePermissions[i].category ? 'bundles' : 'default';
-            if(!section){
-                section = "default";
-            }
             if (!itemsPerSection[section]) {
                 itemsPerSection[section] = [];
             }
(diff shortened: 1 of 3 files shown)


src/Controller/Admin/UserController.php
```diff
@@ -441,7 +441,7 @@ public function getAction(Request $request): JsonResponse
         // get available permissions
         $availableUserPermissionsList = new User\Permission\Definition\Listing();
         $availableUserPermissionsList->setOrderKey('category');
-        $availableUserPermissions = $availableUserPermissionsList->load();
+        $availableUserPermissions = $availableUserPermissionsList->load();$
```
blankse:
```suggestion
        $availableUserPermissions = $availableUserPermissionsList->load();
```

public/js/pimcore/settings/user/role/settings.js
```diff
@@ -72,24 +72,34 @@ pimcore.settings.user.role.settings = Class.create({
         var itemsPerSection = [];
         var sectionArray = [];
         for (var i = 0; i < this.data.availablePermissions.length; i++) {
-            let section = this.data.availablePermissions[i].category;
+            let section = this.data.availablePermissions[i].category ? 'bundles' : 'default';
             if(!section){
                 section = "default";
             }
```
martineiber: We could remove this. `section` cannot be empty because of the line above it. 
```suggestion
```

martineiber (changes requested):

martineiber (approved):

PR #82, opened 2023-05-12
Fix 2FA Setup on user account

### LINKED ISSUE #81: [Bug][Pimcore 11] User cannot setup 2FA
### Pimcore version

11

### Steps to reproduce

- Open users profile
- click on `Setup Two Factor Authentication`
image

--> user gets logged out.

### Expected / actual behavior

_No response_

dvesh3:
Resolves #81


kingjia90 (changes requested):
LGTM just need to add the translation to 
image

`2fa_wrong` (when typing a wrong code) and can be merged

see also https://github.com/pimcore/admin-ui-classic-bundle/blob/85f744fcb92e77f48993bfdf24b77aac7451c0ac/src/Controller/Admin/LoginController.php#L373

PR #83, opened 2023-05-12
[Task]: Defer twemoji variants icons

### LINKED ISSUE #85: [Bug][Pimcore 11] Opening Icon Library kills bowser
### Pimcore version

11

### Steps to reproduce

open icon library, then browser freezes or becomes mega slow

tested with firefox and chrome

### Expected / actual behavior

_No response_

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/85

Kind of work, at least doesn't crash anymore
eg. pressing this knee, it would display all the skin tones
image

but is not coveering cases like this where the basename have more dashes. like the 🫱 and handshake🤝 have the same prefix.

image

### DIFF
--- a/src/Twig/Extension/AdminExtension.php
+++ b/src/Twig/Extension/AdminExtension.php
@@ -51,6 +51,7 @@ class AdminExtension extends AbstractExtension
     {
         return [
             new TwigFilter('pimcore_inline_icon', [$this, 'inlineIcon']),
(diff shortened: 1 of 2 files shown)


martineiber (approved):
Works fine. At least it doesn't kill the browser anymore. For future improvements we should consider some kind of "lazy loading" approach and not load a bunch of icons at once.

PR #115, opened 2023-06-06
Feature/add support for an in filter operator to grid helper

mfedak-notebooksbillger:
### Changes in this pull request

Add support for an 'in' filter operator when filtering numerical fields in the grid.

#### Additional info

At Notebooksbilliger we had a need to filter a column in the grid with multiple values. We have created our own bundle to achieve this but we did have to modify some core code. We are currently changing this core file using a composer patch but would love to see this extra operator check included in future.

Example: https://beta.im.ge/i/in-filter-numeric-columns.FM5r79

(Is related to https://github.com/pimcore/pimcore/pull/15315)

### DIFF
--- a/public/js/pimcore/overrides.js
+++ b/public/js/pimcore/overrides.js
@@ -1126,7 +1126,7 @@ Ext.define('Ext.local.grid.filters.filter.TriFilter', {
             operator: 'in',
             type: 'numeric',
             value: (!stateful && value && value.in) || null
-        }, 'eq');
+        }, 'in');
         me.filter = filter;
         if (me.active) {
             me.setColumnActive(true);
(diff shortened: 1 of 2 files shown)


aryaantony92:
To be merged after `pimcore/pimcore:v1.1.0` is released

mfedak-notebooksbillger:
> To be merged after `pimcore/pimcore:v1.1.0` is released

After pimcore/pimcore:v11.1.0

aryaantony92:
@mfedak-notebooksbillger Looks Good. But could you please refactor the code in override.js such that we don't need to take care of the existing options like gt,lt,eq etc. Like we do for Quantity Value using the inheritance here https://github.com/pimcore/admin-ui-classic-bundle/blob/c4b0b48db3c047f3bc646d6959c84eb2f274171b/public/js/pimcore/object/tags/quantityValue.js#L264

https://github.com/pimcore/admin-ui-classic-bundle/blob/c4b0b48db3c047f3bc646d6959c84eb2f274171b/public/js/pimcore/object/tags/quantityValue.js#L266

aryaantony92:
@mfedak-notebooksbillger Thank you for the PR. We will focus on the refactoring https://github.com/pimcore/admin-ui-classic-bundle/pull/115#issuecomment-1723562449 in a follow up PR.

PR #123, opened 2023-06-15
POEditor translations import

brusch:


robertSt7:
see https://github.com/pimcore/admin-ui-classic-bundle/pull/132

PR #125, opened 2023-06-15
[Data objects] Prevent error when opening object of a class without data object fields

BlackbitDevs:
Steps to reproduce:
1. Create data object class without any fields nor layout components
2. Create an object of this class

Error `Argument #1 ($layout) must be of type array, null given` will get thrown.

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectController.php
+++ b/src/Controller/Admin/DataObject/DataObjectController.php
@@ -585,7 +585,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
         throw $this->createAccessDeniedHttpException();
     }
 
-    private function injectValuesForCustomLayout(array &$layout): void
+    private function injectValuesForCustomLayout(?array &$layout): void
     {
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -585,7 +585,7 @@ static function (Task $task) {
         throw $this->createAccessDeniedHttpException();
     }
 
-    private function injectValuesForCustomLayout(array &$layout): void
+    private function injectValuesForCustomLayout(?array &$layout): void
     {
         foreach ($layout['children'] as &$child) {
```
blankse: When $layout is null this will introduce a warning, right?
```suggestion
        foreach ($layout['children'] ?? [] as &$child) {
```

robertSt7:
@BlackbitDevs Hi, sorry I am not able to reproduce the bug. Could you please provide the Class Definition? Thanks

BlackbitDevs:
Ah seems to have been fixed already in Pimcore 11.

PR #129, opened 2023-06-20
Refactor Tree node config method to service

aryaantony92:


src/Controller/Traits/DocumentTreeConfigTrait.php
```diff
@@ -1,146 +0,0 @@
...
-use Pimcore\Model\Document;
-use Pimcore\Model\Element\ElementInterface;
-use Pimcore\Model\Site;
-use Pimcore\Tool\Admin;
-use Pimcore\Tool\Frontend;
-
-/**
- * @internal
- */
-trait DocumentTreeConfigTrait
```
dvesh3, line 30: this needs to be removed as well https://github.com/pimcore/pimcore/blob/11.x/bundles/SeoBundle/src/Controller/Traits/DocumentTreeConfigWrapperTrait.php

dvesh3: or we keep the `DocumentTreeConfigTrait` as deprecated and use it internally until core 11.1 is released. Otherwise you can't use `pimcore/pimcore:11.0.x` with `admin-ui-classic-bundle:11.1`. wdyt?

dvesh3: reverted here https://github.com/pimcore/admin-ui-classic-bundle/pull/129/commits/39e6499cf619389cc9ea026a194482f8a3bec91d

src/Service/ElementService.php
```diff
@@ -39,4 +66,299 @@ public static function getCustomViewById(string $id): ?array
...
+    /**
+     * @throws \Exception
+     */
+    public function assignDocumentTreeConfig(
+        Document $element,
+        array &$tmpNode,
+        UserProxy|User|null $user
+    ): void
+    {
+        $hasChildren = $element->getDao()->hasChildren(null, Admin::getCurrentUser());
```
kingjia90: I have some doubt on having the `$user` parameter passed around by all these functions (eg. `getElementTreeNodeConfig`), probably would be more easier and "safer" to refactor and just use `Admin::getCurrentUser()` 
See https://github.com/pimcore/admin-ui-classic-bundle/pull/129/files?w=1#r1239506087
Also consider https://github.com/pimcore/static-resolver-bundle

src/Service/ElementService.php
```diff
@@ -39,4 +66,299 @@ public static function getCustomViewById(string $id): ?array
...
+
+    /**
+     * @throws \Exception
+     */
+    public function getElementTreeNodeConfig(
+        ElementInterface $element,
+        UserProxy|User|null $user
+    ): array
+    {
+        $tmpNode = [
```
kingjia90:
```suggestion
        ElementInterface $element
    ): array
    {
        $user = Admin::getCurrentUser();
        $tmpNode = [
```

kingjia90 (commented):
Rest LGTM, just this input/feedback about `$user`

dvesh3 (approved):

src/Service/ElementService.php
dvesh3: Replaced it with [user loader](https://github.com/pimcore/admin-ui-classic-bundle/pull/129/commits/051bec17723cea0171d282d48fd4af244734c352)

src/Controller/Traits/DocumentTreeConfigTrait.php
```diff
@@ -16,21 +16,30 @@
...
 {
     use AdminStyleTrait;
 
+    protected ElementServiceInterface $elementService;
+
+    /**
+     * @param ElementServiceInterface $elementService
+     */
+    #[Required]
+    public function setElementService(ElementServiceInterface $elementService): void
```
kingjia90, line 38: `setElementService` is unused, looks like some leftover after refactoring

https://github.com/pimcore/admin-ui-classic-bundle/commit/4e547a79f37a2f0d531f09d6d386e8887fe30aa7#diff-2419e7874923e300cdf352f832999d7d1d85a77cb69e01b25b1bfa93bcf95c15L37

src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -72,6 +72,11 @@ class DataObjectController extends ElementControllerBase implements KernelContro
 
     private array $classFieldDefinitions = [];
 
+    public function __construct(
+        protected ElementServiceInterface $elementService
+    ) {
+    }
```
kingjia90: could all these be moved under `ElementControllerBase` and use a `getElementService()`?

src/Controller/Traits/DocumentTreeConfigTrait.php
```diff
@@ -16,21 +16,30 @@
...
 use Pimcore\Model\Element\ElementInterface;
-use Pimcore\Model\Site;
-use Pimcore\Tool\Admin;
-use Pimcore\Tool\Frontend;
+use Symfony\Contracts\Service\Attribute\Required;
 
 /**
  * @internal
+ *
+ * @deprecated
```
kingjia90:
```suggestion
 * @deprecated Use elementService instead.
```

src/Service/ElementService.php
```diff
@@ -15,20 +17,46 @@
...
+use Pimcore\Security\User\UserLoader;
+use Pimcore\Tool\Frontend;
+use Pimcore\Bundle\AdminBundle\Controller\Traits\AdminStyleTrait;
+
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
-class ElementService
+/**
+ * @internal
+ */
```
kingjia90, line 38: Not sure, can we just turn into `internal` without a warning period eg. `deprecation`?

kingjia90, line 38: Not sure if that is totally safe change, could be that somebody was exteding it despite it has only 1 internal method`getCustomViewById`?

kingjia90 (changes requested):
LGTM, just some minor things

src/Controller/Traits/DocumentTreeConfigTrait.php
dvesh3: this is used here https://github.com/pimcore/admin-ui-classic-bundle/blob/051bec17723cea0171d282d48fd4af244734c352/src/Controller/Traits/DocumentTreeConfigTrait.php#L52

the service needs to be injected somehow so we use required attribute

kingjia90: ahh, right, i didn't "recognized" the setter injection

src/Controller/Admin/DataObject/DataObjectController.php
dvesh3: yes, good point. moved it 😊

src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -23,7 +23,7 @@
 use Pimcore\Bundle\AdminBundle\Event\ElementAdminStyleEvent;
 use Pimcore\Bundle\AdminBundle\Helper\GridHelperService;
 use Pimcore\Bundle\AdminBundle\Security\CsrfProtectionHandler;
-use Pimcore\Bundle\AdminBundle\Service\ElementService;
+use Pimcore\Bundle\AdminBundle\Service\ElementServiceInterface;
```
kingjia90, line 26:
```suggestion

```

src/Controller/Admin/Document/DocumentController.php
```diff
@@ -19,11 +19,10 @@
 use Imagick;
 use Pimcore;
 use Pimcore\Bundle\AdminBundle\Controller\Admin\ElementControllerBase;
-use Pimcore\Bundle\AdminBundle\Controller\Traits\DocumentTreeConfigTrait;
 use Pimcore\Bundle\AdminBundle\Controller\Traits\UserNameTrait;
 use Pimcore\Bundle\AdminBundle\Event\AdminEvents;
 use Pimcore\Bundle\AdminBundle\Event\ElementAdminStyleEvent;
-use Pimcore\Bundle\AdminBundle\Service\ElementService;
+use Pimcore\Bundle\AdminBundle\Service\ElementServiceInterface;
```
kingjia90, line 25:
```suggestion
```

src/Controller/Admin/Asset/AssetController.php
```diff
@@ -23,7 +23,7 @@
 use Pimcore\Bundle\AdminBundle\Event\ElementAdminStyleEvent;
 use Pimcore\Bundle\AdminBundle\Helper\GridHelperService;
 use Pimcore\Bundle\AdminBundle\Security\CsrfProtectionHandler;
-use Pimcore\Bundle\AdminBundle\Service\ElementService;
+use Pimcore\Bundle\AdminBundle\Service\ElementServiceInterface;
```
kingjia90, line 26:
```suggestion
```

PR #137, opened 2023-06-22
Many-to-many-relations: support multi-file-upload

BlackbitDevs:
For (advanced) many-to-many relations with assets being allowed you can directly upload assets from the relation view:
Bildschirmfoto 2023-06-22 um 16 09 09

Currently this only supports single-file uploads. The same applies under assets context menu for "Upload (Compatibility mode)". With this PR all those places now support multi-file upload.

### DIFF
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1053,7 +1053,7 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                 });
                 win.show();
 
-                var finishedErrorHandler = function (e) {
+                let finishedErrorHandler = function (e) {
                     this.activeUploads--;
                     win.remove(pbar);
 
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1103,7 +1103,6 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                         var data = JSON.parse(request.responseText);
                         if(ev.currentTarget.status < 400 && data.success === true) {
                             success(res);
-                            uploadWindowCompatible.close();
                         } else {
                             failure(res);
                             finishedErrorHandler();
@@ -1124,7 +1123,10 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                 });
 
                 window.setTimeout(function () {
-                    win.close();
+                    if (activeUploads == filesCount) {
+                        win.close();
+                        uploadWindowCompatible.close();
+                    }
                 }.bind(this), 1000);
             },
             afterrender:function(cmp){
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1100,7 +1100,7 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                     };
 
                     let successWrapper = function (ev) {
-                        var data = JSON.parse(request.responseText);
+                        const data = JSON.parse(request.responseText);
                         if(ev.currentTarget.status < 400 && data.success === true) {
                             success(res);
                             if (activeUploads == filesCount) {
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1081,7 +1081,7 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                     win.updateLayout();
 
                     activeUploads++;
-                    let percentComplete = activeUploads / filesCount;
+                    const percentComplete = activeUploads / filesCount;
                     let progressText = file.name + " ( " + Math.floor(percentComplete * 100) + "% )";
                     if (percentComplete == 1) {
                         progressText = file.name + " " + t("please_wait");
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1025,8 +1025,8 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
 
     if (description) {
         items.push({
-           xtype: 'displayfield',
-           value: description
+            xtype: 'displayfield',
+            value: description
         });
     }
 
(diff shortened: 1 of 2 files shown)


robertSt7:
@BlackbitDevs Could you please replace `var` with `let` and `const`. Thank you

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
         buttonConfig: {
             iconCls: 'pimcore_icon_upload'
         },
         listeners: {
             change: function (fileUploadField) {
-                if(fileUploadField.fileInputEl.dom.files[0].size > pimcore.settings["upload_max_filesize"]) {
-                    pimcore.helpers.showNotification(t("error"), t("file_is_bigger_that_upload_limit") + " " + fileUploadField.fileInputEl.dom.files[0].name, "error");
-                    return;
-                }
+                var win = new Ext.Window({
```
kingjia90: Please apply the suggested changes on https://github.com/pimcore/admin-ui-classic-bundle/pull/137#issuecomment-1917028900

The rest LGTM, Thank you in advance

kingjia90 (changes requested):

kingjia90:
One minor graphical thing is that the popup that appears when clicking on the upload button, it doesn't hint that you can select one or more files 
image

public/js/pimcore/helpers.js
BlackbitDevs: @kingjia90 Done. Sorry, was several weeks on sick-leave.

BlackbitDevs:
> One minor graphical thing is that the popup that appears when clicking on the upload button, it doesn't hint that you can select one or more files

Actually people will assume this as they know it from other applications, won't they?
Of course we could change the placeholder to "Select files" - this should include one or multiple. WDYT?

kingjia90:
Yes, that's right, the text change would be fine to me

BlackbitDevs:
Changed to "Select files" - the non-english translations are still maintained outside of github, right?

kingjia90:
Yes, but i would use a new translation key instead.

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
         emptyText: t("select_a_file"),
```
kingjia90:
```suggestion
        emptyText: t("select_files"),
```

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
+                let activeUploads = 0;
+                let filesCount = fileUploadField.fileInputEl.dom.files.length;
+
+                Ext.each(fileUploadField.fileInputEl.dom.files, function (file) {
+                    if (file.size > pimcore.settings["upload_max_filesize"]) {
+                        pimcore.helpers.showNotification(t("error"), t("file_is_bigger_that_upload_limit") + " " + file.name, "error");
+                        return;
+                    }
+
+                    var pbar = new Ext.ProgressBar({
```
kingjia90:
```suggestion
                    let pbar = new Ext.ProgressBar({
```

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
+                    data.append("filename", file.name);
+                    data.append("csrfToken", pimcore.settings['csrfToken']);
+
+                    let request = new XMLHttpRequest();
+                    let res = {
+                        'response': request
+                    };
+
+                    let successWrapper = function (ev) {
+                        var data = JSON.parse(request.responseText);
```
kingjia90:
```suggestion
                        const data = JSON.parse(request.responseText);
```

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
+                    this.activeUploads--;
+                    win.remove(pbar);
+
+                    if(this.activeUploads < 1) {
+                        win.close();
                     }
+                }.bind(this);
+
+                let activeUploads = 0;
+                let filesCount = fileUploadField.fileInputEl.dom.files.length;
```
kingjia90:
```suggestion
                const filesCount = fileUploadField.fileInputEl.dom.files.length;
```

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
         buttonConfig: {
             iconCls: 'pimcore_icon_upload'
         },
         listeners: {
             change: function (fileUploadField) {
-                if(fileUploadField.fileInputEl.dom.files[0].size > pimcore.settings["upload_max_filesize"]) {
-                    pimcore.helpers.showNotification(t("error"), t("file_is_bigger_that_upload_limit") + " " + fileUploadField.fileInputEl.dom.files[0].name, "error");
-                    return;
-                }
+                let win = new Ext.Window({
```
kingjia90:
```suggestion
                const win = new Ext.Window({
```

public/js/pimcore/helpers.js
```diff
@@ -1035,34 +1035,100 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
...
+                        width:465,
+                        text: file.name,
+                        style: "margin-bottom: 5px"
+                    });
+
+                    win.add(pbar);
+                    win.updateLayout();
+
+                    activeUploads++;
+                    let percentComplete = activeUploads / filesCount;
```
kingjia90:
```suggestion
                    const percentComplete = activeUploads / filesCount;
```

translations/admin.en.yaml
```diff
@@ -139,7 +139,7 @@ this_element_cannot_be_edited: 'This element cannot be edited'
 details: Details
 cannot_save_object_please_try_to_edit_the_object_in_detail_view: '<b>Cannot save object!</b><br />Please try to edit the object in the detail view.'
 size: Size
-select_a_file: 'Select a file'
+select_a_file: 'Select files'
```
kingjia90:
```suggestion
select_a_file: 'Select a file'
select_files: 'Select files'
```

kingjia90 (changes requested):

PR #141, opened 2023-06-23
Fix/time field container cut off

### LINKED ISSUE #12446: [Bug]: field container cutting of select
### Expected behavior

Adding a field container to horizontally align any kind of input fields does not cut off the fields.

### Actual behavior

The fields are cut off:

image

image

### Steps to reproduce

1. Open Pimcore Demo
2. Edit any class definition as seen in the screenshot 
3. Reload a related object

> Note that the field container has a label width set to 180. The issue **does not** occur without a label width.

image

torqdev:
Added logic to expose the width element and fix the time drop down from being cutoff. In this case it wasn't the field container that was cutting off the input fields, it was time field drop down that was being cut off due to a set width. Copied the logic used in the select dropdown to change the width based on whether the field container has a value in the label width.

Original Issue: https://github.com/pimcore/pimcore/issues/12446

### DIFF
--- a/public/js/pimcore/object/classes/data/time.js
+++ b/public/js/pimcore/object/classes/data/time.js
@@ -49,6 +49,10 @@ pimcore.object.classes.data.time = Class.create(pimcore.object.classes.data.data
 
         $super();
         this.specificPanel.removeAll();
+
+        var widthItems = this.getWidthPanelItems(this.datax);
+        this.specificPanel.add(widthItems);
+
         var specificItems = this.getSpecificPanelItems(this.datax);
         this.specificPanel.add(specificItems);
 
@@ -134,6 +138,15 @@ pimcore.object.classes.data.time = Class.create(pimcore.object.classes.data.data
         return specificItems;
 
 
+    },
+
+    getWidthPanelItems: function (datax) {
+        return [{
+            xtype: "textfield",
+            fieldLabel: t("width"),
+            name: "width",
+            value: datax.width
+        }];
     },
 
     applySpecialData: function(source) {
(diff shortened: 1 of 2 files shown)


reilly-webster:

torqdev:
recheck

reilly-webster:
recheck

public/js/pimcore/object/classes/data/time.js
```diff
@@ -49,6 +49,10 @@ pimcore.object.classes.data.time = Class.create(pimcore.object.classes.data.data
 
         $super();
         this.specificPanel.removeAll();
+
+        var widthItems = this.getWidthPanelItems(this.datax);
```
robertSt7:
```suggestion
        const widthItems = this.getWidthPanelItems(this.datax);
```

public/js/pimcore/object/tags/time.js
```diff
@@ -29,18 +29,28 @@ pimcore.object.tags.time = Class.create(pimcore.object.tags.abstract, {
     },
 
     getLayoutEdit: function () {
-        this.component = new Ext.form.TimeField({
+        const options = {
             fieldLabel: this.fieldConfig.title,
             format: "H:i",
             emptyText: "",
-            width: 200,
+            width: this.fieldConfig.width ? this.fieldConfig.width : 250,
```
robertSt7:
```suggestion
            width: this.fieldConfig.width ? this.fieldConfig.width : 200,
```

robertSt7:
@torqdev @reilly-webster Thanks for the fix :)

PR #145, opened 2023-07-03
Don't use session if request is stateless

dpfaffenbauer:


dpfaffenbauer:
thank you CLA Bot, but there is no link to sign :D

fashxp:
@dpfaffenbauer see cla bot's message ;-)
> You can sign the CLA by just posting a Pull Request Comment same as the below format.

dpfaffenbauer:

dpfaffenbauer:
@fashxp thanks :)

blankse:
The target branch should be `1.0` if it is a bug.

dpfaffenbauer:
@blankse done

src/EventListener/AdminSessionBagListener.php
```diff
@@ -49,6 +49,10 @@ public function onKernelRequest(RequestEvent $event): void
             return;
         }
 
+        if ($event->getRequest()->attributes->get('_stateless', false)) {
```
dvesh3, line 52: we need to handle this in a generic way (maybe by adding abstract in the core) as there are similar implementations e.g. https://github.com/pimcore/ecommerce-framework-bundle/blob/4c2f9f5f20d128a1f61ef0a5335516c0a8cba1c5/src/EventListener/SessionBagListener.php#L58

https://github.com/pimcore/personalization-bundle/blob/e27b008210a9a8be01cedf770bafd7bd3f63e727/src/Targeting/EventListener/TargetingSessionBagListener.php#L69

And also there could be custom ones https://github.com/pimcore/pimcore/blob/82b7f2d53448694aa6d1e42d5c3e92dfa8722117/doc/19_Development_Tools_and_Details/35_Working_with_Sessions.md#register-a-session-bag-through-eventlistener

dpfaffenbauer: What generic way do you imagine? it is listening to the generic Symfony `KernelEvents::REQUEST` event and adds the session bag to the generic Symfony Session

dvesh3: I mean to handle checks like this in a central place:
```
if (!$event->isMainRequest()) {
    return;
}

if ($event->getRequest()->attributes->get('_stateless', false)) {
    return;
}
```
since this needs to be changed in other implementations as well.

dvesh3: However this is something we should do in minor releases as the core package requirement needs to bumped so let's keep it like this for now.

dvesh3:
@dpfaffenbauer thanks!

PR #147, opened 2023-07-04
Fix xss in admin login 2fa setup page

aryaantony92:


src/Controller/Admin/LoginController.php
```diff
@@ -355,7 +355,7 @@ public function twoFactorSetupAuthenticationAction(
         $proxyUser = $this->getAdminUser(true);
 
         if ($request->query->get('error')) {
-            $params['error'] = $request->query->get('error');
+            $params['error'] = htmlspecialchars($request->query->get('error'));
```
dvesh3: I guess instead of escaping special chars here, we should re-generate a csrf token like lost pass action https://github.com/pimcore/admin-ui-classic-bundle/blob/4e537d7713f18a51b3acbe321f67f58827f2ecc3/src/Controller/Admin/LoginController.php#L261

since we already handle this on the view https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/templates/admin/login/two_factor_setup.html.twig#L27

aryaantony92: This didn't work and the error message was retrieved from the request.

PR #148, opened 2023-07-04
Handle pimcore.helpers.download exceptions

### LINKED ISSUE #126: Handle pimcore.helpers.download exceptions
At the moment, exceptions are thrown e.g. `createNotFoundException` and not handled in js.

this one tricky to reproduce with default scenarios so it needs a bit of a tweaking:
 - tweak a controller action (for downloading) to return `createNotFoundException` e.g. AssetController::downloadAction
 - Open asset image and click download.
 - the request will be redirected to 404 page instead of handling it gracefully in admin UI.

robertSt7:
resolves #126

### DIFF
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -884,12 +884,10 @@ pimcore.helpers.download = function (url) {
     }
     iframe.src = url;
 
-    document.getElementById('download_helper_iframe').onload = function() {
-        if(iframe.src !== 'about:blank') {
-            const title = iframe.contentDocument.title;
-            pimcore.helpers.showNotification(t('error'), title, 'error');
-            iframe.src = 'about:blank';
-        }
+    iframe.onload = function() {
+          const title = iframe.contentDocument.title;
+          pimcore.helpers.showNotification(t('error'), title, 'error');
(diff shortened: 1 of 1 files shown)


public/js/pimcore/helpers.js
```diff
@@ -876,7 +876,21 @@ pimcore.helpers.download = function (url) {
...
+    }
+    iframe.src = url;
+
+    document.getElementById('download_helper_iframe').onload = function() {
+        if(iframe.src !== 'about:blank') {
+            const title = iframe.contentDocument.title;
+            pimcore.helpers.showNotification(t('error'), title, 'error');
+            iframe.src = 'about:blank';
+        }
+    }
```
kingjia90:
```suggestion
    iframe.onload = function() {
          const title = iframe.contentDocument.title;
          pimcore.helpers.showNotification(t('error'), title, 'error');
          iframe.src = 'about:blank';
    }
```
this may doing the trick in a simpler way.
When the download is triggered, there's no DOM `onload` event at all

kingjia90: let's also add a js comment on the purpose of this `iframe`

PR #153, opened 2023-07-07
Fix Base Units in QuantityValue Unit Definition GridView

hbauer:
I have created 196 QuantityUnits by api and the value in the column 'Base Unit' column is incorrect. Additionally, the dropdown menu for selecting a Base Unit will only display 25 items.

This issue arises from the restriction of 25 items in the function _unitProxyGetAction_ in the _QuantityValueController_. Although this restriction doesn't affect the paging of the grid view, it limits the _baseUnitStore_ used in _unitsettings.js_ to always contain only the first 25 QuantityUnits. As a result, the Base Units may not be found in some cases and will be displayed as (Empty).

You can see this in the attached screenshots below.

Current Behaviour:
Pimcore 11 - QuantityValue Unit Definition - Wrong

Expected Behaviour:
Pimcore 11 - QuantityValue Unit Definition - Ok

Testfile for Import:
[quantityvalue_units.sql.zip](https://github.com/pimcore/admin-ui-classic-bundle/files/11980815/quantityvalue_units.sql.zip)

### DIFF
--- a/src/Controller/Admin/DataObject/QuantityValueController.php
+++ b/src/Controller/Admin/DataObject/QuantityValueController.php
@@ -64,8 +64,8 @@ class QuantityValueController extends AdminAbstractController
         $list->setOrder($order);
         $list->setOrderKey($orderKey);
 
-        $list->setLimit((int)$request->get('limit', 25));
-        $list->setOffset((int)$request->get('start', 0));
+        $list->setLimit((int)$request->get('limit'));
+        $list->setOffset((int)$request->get('start'));
 
         $condition = '1 = 1';
         if ($request->get('filter')) {


kingjia90:
Thank you!

I've opened a new PR on main repository, because the issue would be affecting Pimcore version 10.x
https://github.com/pimcore/pimcore/pull/15576

Could you please verify that the other PR works also fine? I find the default fallback is useful for the grid itself, but not when loading the dropdown, therefore the different suggested approach. 
Not sure if it could be considered as a potential behaviour change that leads to bc break, while the other PR is directly fixing the dropdown

kingjia90:
Thank you again for the PR but closing in favor of https://github.com/pimcore/pimcore/pull/15576

PR #160, opened 2023-07-13
Open object with incompatible latest version

BlackbitDevs:
Steps to reproduce:

1. Create data object class with input field `test`
2. Create object of this class and enter "test" into field `test`
3. Either an `autoSave` draft version gets created or click "Save draft"
4. Convert field `test` in the class definition to a many-to-one relation
5. Reload the object

Result: There will be the error `Pimcore\Model\DataObject\ClassDefinition\Data\ManyToOneRelation::preGetData(): Return value must be of type ?Pimcore\Model\Element\ElementInterface, string returned`

Cause: When opening a data object, the latest version gets loaded in https://github.com/pimcore/admin-ui-classic-bundle/blob/7425a0782b1697550b4019d31f8ad52e5fcbbf0a/src/Controller/Admin/DataObject/DataObjectController.php#L311
As this contains serialized data, this data may not be compatible anymore to the changed field types. Before the field getters enforced return types this was no problem but nowadays the getters do only allow `null` or the correct return value type.

Solution: If the latest version is not compatible, the object will get loaded from database.

### DIFF
--- a/public/js/pimcore/object/tags/wysiwyg.js
+++ b/public/js/pimcore/object/tags/wysiwyg.js
@@ -121,7 +121,6 @@ pimcore.object.tags.wysiwyg = Class.create(pimcore.object.tags.abstract, {
         this.component.on("afterrender", function() {
             Ext.get(this.editableDivId).dom.setAttribute("contenteditable", "false");
         }.bind(this));
-        this.component.disable();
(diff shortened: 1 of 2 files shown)


robertSt7:
@BlackbitDevs Thanks for the PR

PR #174, opened 2023-07-20
Make BundleManagerEvents extend from \Pimcore\Event\BundleManagerEvents

### LINKED ISSUE #172: Are BundleManagerEvents really needed in that bundle
see https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/src/Event/BundleManagerEvents.php

Same file also still exists in pimcore/pimcore, see https://github.com/pimcore/pimcore/blob/11.x/lib/Event/BundleManagerEvents.php, and is also used in PimcoreBundleManager, which lives in pimcore/pimcore

dvesh3:
Resolves #172

### DIFF
--- a/src/Event/BundleManagerEvents.php
+++ b/src/Event/BundleManagerEvents.php
@@ -16,41 +16,9 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\AdminBundle\Event;
 
-final class BundleManagerEvents
+/**
+ * @TODO this class is only here for BC reasons and should be removed in Pimcore 12
+ */
+final class BundleManagerEvents extends \Pimcore\Event\BundleManagerEvents
 {
-    /**
-     * The CSS_PATHS event is triggered for paths to CSS files which are about to be loaded for the admin interface.
-     *
-     * @Event("Pimcore\Bundle\AdminBundle\Event\BundleManager\PathsEvent")
-     *
-     * @var string
-     */
-    const CSS_PATHS = 'pimcore.bundle_manager.paths.css';
-
-    /**
-     * The JS_PATHS event is triggered for paths to JS files which are about to be loaded for the admin interface.
-     *
-     * @Event("Pimcore\Bundle\AdminBundle\Event\BundleManager\PathsEvent")
-     *
-     * @var string
-     */
-    const JS_PATHS = 'pimcore.bundle_manager.paths.js';
-
-    /**
-     * The EDITMODE_CSS_PATHS event is triggered for paths to CSS files which are about to be loaded in editmode.
-     *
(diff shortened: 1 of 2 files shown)


dvesh3:
the tests will pass on `pimcore/pimcore` v11.0.4 release.

src/Event/BundleManagerEvents.php
```diff
@@ -16,41 +16,9 @@
 
 namespace Pimcore\Bundle\AdminBundle\Event;
 
-final class BundleManagerEvents
+/**
+ * @TODO this class is only here for BC reasons and should be removed in Pimcore 12
```
blankse, line 20: Shouldn't it be deprecated then?
```suggestion
 * @deprecated this class is only here for BC reasons and should be removed in Pimcore 12
```

dvesh3: we want to avoid this intentionally as this was already deprecated on 10.6. https://github.com/pimcore/pimcore/pull/15468/files#diff-29364e2d0b002e03d5bb2e7e6b0b6c255943ddf7926d965527d01b8c48509c5eL19

blankse: Its bad yes. But I think the developer should know that now he should use the one in the core, right? So the deprecated annotation would help.

dvesh3: It will stay there for v11 version so shouldn't be a problem to keep using it.

kingjia90: good point but it cannot be deprecated in a bug fix release, it likely would need the deprecated tag in the latest minor (like what happened for `10.6`) before major, right?

blankse: Why not adding it in 11.1? So the developer have more time to change it.

jdreesen: Why can't it be deprecated in a bug fix release?

dvesh3: because bugfixes are not suppose the change the running behavior of the application. we want to merge it in bugfix to avoid confusion with https://github.com/pimcore/pimcore/pull/15468/files#diff-29364e2d0b002e03d5bb2e7e6b0b6c255943ddf7926d965527d01b8c48509c5eL19

jdreesen: But what does a deprecation change (except saying: please don't use me anymore)?

dvesh3: the notice changes the context for running code...right?

jdreesen: Hmm, that might be true for `trigger_deprecation()`, yes :thinking: 
But adding a `@deprecated` in the PHPDoc shouldn't change anything...
Anyway, I'm fine with deprecating it in `11.1`.

PR #177, opened 2023-07-20
Fix deprecation in asset/document tree

blankse:
```
PHP Deprecated: Since pimcore/pimcore 11.0: Passing id as string to method Pimcore\Model\Asset::getById is deprecated
```
```
PHP Deprecated: Since pimcore/pimcore 11.0: Passing id as string to method Pimcore\Model\Document::getById is deprecated
```

### DIFF
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -259,7 +259,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
 
         $assets = [];
         $cv = [];
-        $asset = Asset::getById($allParams['node']);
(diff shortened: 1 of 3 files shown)


kingjia90 (changes requested):
LGTM, can you also add this 
https://github.com/blankse/admin-ui-classic-bundle/blob/ee662bf80b52982191ed7e946ea6aa58ee86b91c/src/Helper/GridHelperService.php#L681 to the PR?

blankse:
@kingjia90 Done https://github.com/pimcore/admin-ui-classic-bundle/pull/177/commits/663184c86954a513e95396ba5a503e3f05720dfb

PR #186, opened 2023-07-31
Added fallback user owner to the system user

aashan10:
Fixes #185

PR #187, opened 2023-07-31
Added fallback system user for  userOwnerName and userModificationName

### LINKED ISSUE #185: Type error when opening an object
The database schema allows `userOwner` field of the `objects` table to be nullable.
https://github.com/pimcore/pimcore/blob/cfa3ba70a0eb5794277a242ebc4396fae7f8b898/bundles/InstallBundle/dump/install.sql#L277

However, if there are any data objects with null values in the database, there is a type error triggered by 

https://github.com/pimcore/admin-ui-classic-bundle/blob/de8fff41a353b9e850e4394bbbb086ba89ad651f/src/Controller/Traits/UserNameTrait.php#L44

on this line 

https://github.com/pimcore/admin-ui-classic-bundle/blob/de8fff41a353b9e850e4394bbbb086ba89ad651f/src/Controller/Admin/DataObject/DataObjectController.php#L413

aashan10:
Resolves #185 
Followup to #186


dvesh3:
Closed in favor of #87. thanks @aashan10

PR #189, opened 2023-07-31
Fix: Create WebsiteSetting without a siteId and a language

robertSt7:
connected to https://github.com/pimcore/pimcore/pull/15658

### DIFF
--- a/public/js/pimcore/settings/website.js
+++ b/public/js/pimcore/settings/website.js
@@ -557,7 +557,9 @@ pimcore.settings.website = Class.create({
         let res = store.add({
             name: key,
             data: value,
-            type: type
+            type: type,
+            language: '',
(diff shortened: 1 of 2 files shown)


public/js/pimcore/settings/website.js
```diff
@@ -557,7 +557,9 @@ pimcore.settings.website = Class.create({
         let res = store.add({
             name: key,
             data: value,
-            type: type
+            type: type,
+            language: '',
+            siteId: 0
```
dvesh3:
```suggestion
            siteId: null
```
`null` => no site, 0 => `main site`

dvesh3 (approved):
Depends on https://github.com/pimcore/pimcore/pull/15658

PR #190, opened 2023-08-01
Support text selection for readonly wysiwyg fields

BlackbitDevs:
Steps to reproduce bug:

1. Create data object class with wysiwyg field
2. Create object
3. Enter some content in wysiwyg field, save object
4. change field definition for the wysiwyg field to be `not editable`
5. Reload object

The content is now not selectable.
With this PR it is selectable (and thus also copyable).

It is similar to what has been implemented in https://github.com/pimcore/pimcore/pull/3693

### DIFF
--- a/public/js/pimcore/object/tags/wysiwyg.js
+++ b/public/js/pimcore/object/tags/wysiwyg.js
@@ -121,7 +121,6 @@ pimcore.object.tags.wysiwyg = Class.create(pimcore.object.tags.abstract, {
         this.component.on("afterrender", function() {
             Ext.get(this.editableDivId).dom.setAttribute("contenteditable", "false");
         }.bind(this));
-        this.component.disable();
         return this.component;
     },
 


kingjia90:
The problem is that it looks different, with disabled it grays out
image

Would rather change the css under `bundles/pimcoreadmin/css/ext-modifications.css`, maybe 
image
an exception for `.object_field_type_wysiwyg`
UPDATE: was wrong, the css is restoring by setting `all!important`, it seems something caused by `x-mask`

kingjia90:
Thank you for the PR! 
Closing in favor of https://github.com/pimcore/pimcore/pull/15720 as this PR is changing behaviour/display (grayed zone) and was actually a potential bug for 10.6

kingjia90:
Closing in favor of https://github.com/pimcore/admin-ui-classic-bundle/pull/206

PR #191, opened 2023-08-01
Support text selection / clickable UI for readonly objectbrick fields

BlackbitDevs:
Similar to #190 when you set an object brick field to `not editable` it gets completely disabled. Thus you can neither select any text nor can you click buttons to jump to relations (or other UI actions).

Without this PR:
Bildschirmfoto 2023-08-01 um 17 36 46

With this PR:
Bildschirmfoto 2023-08-01 um 17 36 19

### DIFF
--- a/public/js/pimcore/object/tags/objectbricks.js
+++ b/public/js/pimcore/object/tags/objectbricks.js
@@ -300,17 +300,12 @@ pimcore.object.tags.objectbricks = Class.create(pimcore.object.tags.abstract, {
                     afterrender: function (childConfig, dataProvider, manuallyAdded, panel) {
                         if (!panel.__tabpanel_initialized) {
                             var copy = Ext.decode(Ext.encode(childConfig));
-                            var children = this.getRecursiveLayout(copy, null, {
+                            var children = this.getRecursiveLayout(copy, this.fieldConfig.noteditable, {
                                 containerType: "objectbrick",
                                 containerKey: type,
                                 ownerName: this.fieldConfig.name,
                                 applyDefaults: manuallyAdded
                             }, false, true, dataProvider);
-                            if (this.fieldConfig.noteditable && children) {
-                                children.forEach(function (record) {
-                                    record.disabled = true;
-                                });
-                            }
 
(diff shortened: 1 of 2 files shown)


kingjia90:
Thank you for the PR! 
Closing in favor of https://github.com/pimcore/pimcore/pull/15720 as this PR is changing behaviour/display (grayed zone) and was actually a potential bug for 10.6

public/js/pimcore/object/tags/objectbricks.js
```diff
@@ -300,17 +300,12 @@ pimcore.object.tags.objectbricks = Class.create(pimcore.object.tags.abstract, {
...
                                 containerType: "objectbrick",
                                 containerKey: type,
                                 ownerName: this.fieldConfig.name,
                                 applyDefaults: manuallyAdded
                             }, false, true, dataProvider);
-                            if (this.fieldConfig.noteditable && children) {
-                                children.forEach(function (record) {
-                                    record.disabled = true;
-                                });
-                            }
```
kingjia90, line 313: to keep the grayed mask, i've added here
```
if (this.fieldConfig.noteditable){
    panel.mask();
}
```
and a css rule as 

```
.pimcore_objectbrick_item.x-masked .x-mask{
    pointer-events: none;
}
```
to make it work without visual changes, could you please also verify? 

Then it should go on 10.6 as well. Let me know if you can do that before the end of the day, because i may have to takeover as tomorrow would be the last 10.6 release

kingjia90: Follow-up note: this suggestion to add a base `mask` is meant to be only valid for 10.6 to avoid the display changes, in 11 it can be considered to be removed https://github.com/pimcore/pimcore/pull/15759#issue-1854317248

kingjia90: the removal of the mask would be postponed a little bit and may take a confirmation first, merging this PR as the equivalent/port of https://github.com/pimcore/pimcore/pull/15759 for Pimcore 11

PR #192, opened 2023-08-02
Update ext-modifications.css

Corepex:
This ticket will fix the padding on tree elements.


blankse:
Target branch should be `1.0`

PR #197, opened 2023-08-03
Fix username is set to unkown when userOwner of userModification is null

bramstroker:
Superseeds https://github.com/pimcore/admin-ui-classic-bundle/pull/187

Imho this fix is better.

Because first of all in the other PR user is set to 0 which is `system` user, which is not correct.
In my fix this is set to unknown which is consistent with the existing behaviour when a certain userId does not exist.

Also the fix is made in the trait which also fixes it for potential other places where the trait is used.


bramstroker:

bramstroker:
recheck

blankse:
Target branch should be `1.0`

bramstroker:
> Target branch should be `1.0`

@blankse See #198. It's based and targeted on correct branch.
Closing this PR.

PR #198, opened 2023-08-03
Fix username is set to unkown when userOwner of userModification is null

### LINKED ISSUE #185: Type error when opening an object

bramstroker:
Superseeds https://github.com/pimcore/admin-ui-classic-bundle/pull/187

Imho this fix is better.

Because first of all in the other PR user is set to 0 which is `system` user, which is not correct.
In my fix this is set to unknown which is consistent with the existing behaviour when a certain userId does not exist.

Also the fix is made in the trait which also fixes it for potential other places where the trait is used.

### DIFF
--- a/src/Controller/Traits/UserNameTrait.php
+++ b/src/Controller/Traits/UserNameTrait.php
@@ -37,7 +37,7 @@ trait UserNameTrait
     }
 
     /**
-     * @param int $userId The User ID.
+     * @param int|null $userId The User ID.
      *
      * @return array{userName: string, fullName: string}
      */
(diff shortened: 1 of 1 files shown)


dvesh3 (approved):

dvesh3:
@bramstroker thanks!

PR #201, opened 2023-08-08
[Bug] Adjust rename function

Corepex:
We should use the same default value for all rename operations ...

### DIFF
--- a/public/js/pimcore/asset/tree.js
+++ b/public/js/pimcore/asset/tree.js
@@ -1184,7 +1184,7 @@
              elementType: "asset",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: record.data.key
+             default: Ext.util.Format.htmlDecode(record.data.key)
          };
          pimcore.elementservice.editElementKey(options);
      },
--- a/public/js/pimcore/asset/tree.js
+++ b/public/js/pimcore/asset/tree.js
@@ -1184,7 +1184,7 @@
              elementType: "asset",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: Ext.util.Format.htmlDecode(record.data.text)
+             default: record.data.key
          };
          pimcore.elementservice.editElementKey(options);
      },
--- a/public/js/pimcore/asset/tree.js
+++ b/public/js/pimcore/asset/tree.js
@@ -1184,7 +1184,7 @@
              elementType: "asset",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: Ext.util.Format.htmlDecode(record.data.key)
+             default: record.data.key
          };
          pimcore.elementservice.editElementKey(options);
      },
--- a/public/js/pimcore/object/tree.js
+++ b/public/js/pimcore/object/tree.js
@@ -1119,7 +1119,7 @@ pimcore.registerNS("pimcore.object.tree");
              elementType: "object",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: Ext.util.Format.htmlDecode(record.data.text)
+             default: record.data.key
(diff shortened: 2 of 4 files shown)


src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1162,8 +1162,11 @@ private function executeUpdateAction(DataObject $object, mixed $values): array
                 return ['success' => false, 'message' => "You can't rename or relocate if there's a newer not published version"];
             }
         }
-        
-        $key = Service::getValidKey($values['key'], 'object') ?? null;
+
+        $key = $values['key'] ?? null;
+        if($key) {
+            $key = Service::getValidKey($values['key'], 'object');
```
blankse:
```suggestion
        if ($key) {
            $key = Service::getValidKey($key, 'object');
```

Corepex: thx

src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1163,6 +1164,9 @@ private function executeUpdateAction(DataObject $object, mixed $values): array
         }
 
         $key = $values['key'] ?? null;
+        if ($key) {
+            $key = Service::getValidKey($key, 'object');
+        }
```
dvesh3, line 1169:
```suggestion
```
just found out that we handle it on client side https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/pimcore/elementservice.js#L459

public/js/pimcore/asset/tree.js
```diff
@@ -1184,7 +1184,7 @@
              elementType: "asset",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: Ext.util.Format.htmlDecode(record.data.text)
+             default: record.data.key
```
dvesh3: please revert as this is required https://github.com/pimcore/pimcore/pull/5206/files#diff-9d9bcbb5d9a089108c6e32c3b3885884382b70ce63550e82ddebbc81d33a39eeR310

public/js/pimcore/object/tree.js
```diff
@@ -1119,7 +1119,7 @@ pimcore.registerNS("pimcore.object.tree");
              elementType: "object",
              elementSubType: record.data.type,
              id: record.data.id,
-             default: Ext.util.Format.htmlDecode(record.data.text)
+             default: record.data.key
```
dvesh3: please revert as this is required https://github.com/pimcore/pimcore/pull/5206/files#diff-9d9bcbb5d9a089108c6e32c3b3885884382b70ce63550e82ddebbc81d33a39eeR310

src/Controller/Admin/DataObject/DataObjectController.php
Corepex: is client-side enough? i guess we should also handle it on the server or?

public/js/pimcore/asset/tree.js
Corepex: done with 7131182ed24c7c1a40acce25e2fabc411a298da7

public/js/pimcore/object/tree.js
Corepex: done with 7131182ed24c7c1a40acce25e2fabc411a298da7

src/Controller/Admin/DataObject/DataObjectController.php
dvesh3: yes, right.

PR #207, opened 2023-08-10
Fix: build ace-editor with webpack

### LINKED ISSUE #110: Ace Editor issues
it seems ace editor does not work

i tried these options:
```
var jsonMode = ace.require("ace/mode/json").Mode;
editor.session.setMode(new jsonMode());
```
and
```
editor.session.setMode("ace/mode/json");
```

and it did not work

Also, are npm packages being built when imported from composer?

robertSt7:
resolves #110


templates/admin/index/index.html.twig
```diff
@@ -186,7 +186,9 @@
     "lib/ext-plugins/portlet/PortalDropZone.js",
     "lib/ext-plugins/portlet/Portlet.js",
     "lib/ext-plugins/portlet/PortalColumn.js",
-    "lib/ext-plugins/portlet/PortalPanel.js"
+    "lib/ext-plugins/portlet/PortalPanel.js",
+    "../build/admin/ace-builds/src-min-noconflict/ace.js",
+    "../build/admin/ace-builds/src-min-noconflict/ext-modelist.js"
```
dvesh3, line 191: getting error in the console now with csp enabled:
```
Refused to create a worker from 'blob:http://localhost/b02787d9-5feb-4b84-82c7-f5389283edf8' because it violates the following Content Security Policy directive: "script-src 'self'`.....
```

dvesh3: The warning is only there on chrome and doesn't have any affect on the working of editor.

dvesh3 (approved):

PR #208, opened 2023-08-11
Update DataObjectActionsTrait.php

djbodya:
This commit fixes error "Pimcore\Model\DataObject\ClassDefinition\Data\ManyToOneRelation::getDataFromGridEditor(): Argument #1 ($data) must be of type array, string given, called in /var/www/html/vendor/pimcore/admin-ui-classic-bundle/src/Controller/Admin/DataObject/DataObjectActionsTrait.php on line 279"


dvesh3:
Regression of https://github.com/pimcore/admin-ui-classic-bundle/pull/204

aryaantony92:
@djbodya Thanks!

PR #222, opened 2023-08-21
ExtJSCommand: Don't override $defaultName property

blankse:
```
The "Symfony\Component\Console\Command\Command::$defaultName" property is considered final. You should not override it in "Pimcore\Bundle\AdminBundle\Command\ExtJSCommand".
```


robertSt7:
@blankse Thanks

PR #223, opened 2023-08-21
[Bug] Event pimcore.events.prepareAffectedNodes details not correctly built after renaming assets or data objects

### LINKED ISSUE #203: [Bug] Event pimcore.events.prepareAffectedNodes details not correctly built after renaming assets or data objects
### Version

1.0 - 1.x

### Steps to reproduce

Pretty hard to without a custom tree, but should be available with an internal feature

### Actual Behavior

After renaming an asset or an data object in the custom tree, the renamed element is marked as dirty even though it is updated in the standard tree correctly

image

Another Problem is, if you try to rename the element again, the old value will show up

### Expected Behavior

Event pimcore.events.prepareAffectedNodes is built correctly and the field is not marked as dirty anymore and updated accordingly as it works for the documents.

robertSt7:
resolves #203

### DIFF
--- a/public/js/pimcore/elementservice.js
+++ b/public/js/pimcore/elementservice.js
@@ -482,7 +482,7 @@ pimcore.elementservice.editObjectKeyComplete = function (options, button, value,
                 var index, record;
                 for (index = 0; index < affectedNodes.length; index++) {
                     record = affectedNodes[index];
-                    pimcore.elementservice.refreshNode(record);
+                    pimcore.elementservice.refreshNode(record.parentNode);
                 }
 
                 try {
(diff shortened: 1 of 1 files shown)


mattamon (approved):
Works now for assets and objects! LGTM

dvesh3:
Needs more discussion as the changes in the PR makes the parent node refresh which resets the siblings state.

robertSt7:
Try to refresh the store for that changed record instead of refreshing the parent.

public/js/pimcore/elementservice.js
```diff
@@ -482,7 +482,7 @@ pimcore.elementservice.editObjectKeyComplete = function (options, button, value,
                 var index, record;
                 for (index = 0; index < affectedNodes.length; index++) {
                     record = affectedNodes[index];
-                    pimcore.elementservice.refreshNode(record);
+                    pimcore.elementservice.refreshNode(record.parentNode);
```
kingjia90, line 485: if the `parentNode` refreshing approach is valid, need to be refactored to gather all the parentNodes in one array, remove duplicates with `array_unique` and refresh this new array, instead of refreshing for each affected node, since these nodes could share the same parents, no need to refresh the exact same parent multiple times.

same goes for Documents https://github.com/pimcore/admin-ui-classic-bundle/blob/e2e658f199a4662915caad5b8260876e9fde75b4/public/js/pimcore/elementservice.js#L434

dvesh3:
Closed in favor of #266

PR #224, opened 2023-08-21
[Thumbnails] Fix download thumbnails action to use auto formats

dvesh3:
Related to pimcore/pimcore#15676

### DIFF
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -1241,6 +1241,13 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         }
 
         if ($thumbnail) {
+            $thumbnailConfig = $thumbnail->getConfig();
+            if ($thumbnailConfig->getFormat() === 'SOURCE' &&
+                $autoFromatConfigs = $thumbnailConfig->getAutoFormatThumbnailConfigs()) {
+                $autoFromatConfig = current($autoFromatConfigs);
+                $thumbnail = $image->getThumbnail($autoFromatConfig);
+            }
+
             $thumbnailFile = $thumbnailFile ?: $thumbnail->getLocalFile();
 
             $downloadFilename = preg_replace(
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -1243,9 +1243,9 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         if ($thumbnail) {
             $thumbnailConfig = $thumbnail->getConfig();
             if ($thumbnailConfig->getFormat() === 'SOURCE' &&
-                $autoFromatConfigs = $thumbnailConfig->getAutoFormatThumbnailConfigs()) {
-                $autoFromatConfig = current($autoFromatConfigs);
-                $thumbnail = $image->getThumbnail($autoFromatConfig);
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/Asset/AssetController.php
```diff
@@ -1241,6 +1241,13 @@ public function downloadImageThumbnailAction(Request $request): BinaryFileRespon
         }
 
         if ($thumbnail) {
+            $thumbnailConfig = $thumbnail->getConfig();
+            if ($thumbnailConfig->getFormat() === 'SOURCE' &&
+                $autoFromatConfigs = $thumbnailConfig->getAutoFormatThumbnailConfigs()) {
```
blankse: `getAutoFormatThumbnailConfigs()` is internal

kingjia90: is there any drawback? i see that the controller itself and other methods used in here are internal too

blankse: We should remove the internal flag if the method is used outside of the repository.
So it can't be changed without BC break, because it is used external.

kingjia90, line 1246: a typo in the var name
```suggestion
                $autoFormatConfigs = $thumbnailConfig->getAutoFormatThumbnailConfigs()) {
```

src/Controller/Admin/Asset/AssetController.php
```diff
@@ -1241,6 +1241,13 @@ public function downloadImageThumbnailAction(Request $request): BinaryFileRespon
...
+                $autoFromatConfig = current($autoFromatConfigs);
```
kingjia90: a typo in the var name

```suggestion
                $autoFormatConfigs = $thumbnailConfig->getAutoFormatThumbnailConfigs()) {
                $autoFormatConfig = current($autoFormatConfigs);
```

src/Controller/Admin/Asset/AssetController.php
```diff
@@ -1241,6 +1241,13 @@ public function downloadImageThumbnailAction(Request $request): BinaryFileRespon
...
+                $thumbnail = $image->getThumbnail($autoFromatConfig);
```
kingjia90:
```suggestion
                $thumbnail = $image->getThumbnail($autoFormatConfig);
```

src/Controller/Admin/Asset/AssetController.php
kingjia90: I think that's a good point, looks like it may potentially and inadvertently change the expected output and be harder to maintain, but don't we just have to make sure that any change to `getAutoFormatThumbnailConfigs` would have the right follow-up to all its usage and "internally" treat it as if something that can bc-break? 🤔 
By removing the internal flag, it kind of "invites" the devs to use it in their own scripts, which doesn't seem a purpose for this method

blankse: @kingjia90 When you need to use it external, why should be a dev not use it external? When you use internal methods in other repositories it will be hard to maintain. Because every change on internal methods you must search in all repositories for usages. Do you realy want handle this? Maybe this logic should be in the core and you make a new external method for this repository and for all devs?

kingjia90: ahh, now i got it, it's because these are in different repositories, looks like that may change everything

dvesh3: this bundle is still tightly coupled with the core as it still uses lot of core internal classes so we can go ahead with this change.

PR #227, opened 2023-08-22
Fix HTML validation of login form

NiklasBr:
Fix most *info*, *warning*, and *errors* reported by: https://validator.w3.org/nu/?showsource=yes&showoutline=yes&doc=https%3A%2F%2Fdemo.pimcore.fun%2Fadmin%2Flogin

### DIFF
--- a/templates/admin/login/login.html.twig
+++ b/templates/admin/login/login.html.twig
@@ -40,12 +40,6 @@
 
     <a id="proceed_with_unsupported_browser" href="#" >{{ 'Click here to proceed'|trans([], 'admin') }}</a>
 </div>
-
-<style type="text/css">
-    #loginform {
-        display: none;
-    }
-</style>
 {% endif %}
 {% endblock %}
 
--- a/templates/admin/login/login.html.twig
+++ b/templates/admin/login/login.html.twig
@@ -1,7 +1,7 @@
 {% extends '@PimcoreAdmin/admin/login/layout.html.twig' %}
 
 {% block content %}
-<div id="loginform">
+<div id="loginform" {% if not browserSupported %}style="display:none;"{% endif %}>
     <form id="form-element" method="post" action="{{ path('pimcore_admin_login_check', {'perspective' : app.request.get('perspective')|striptags}) }}">
 
         {% if error is defined %}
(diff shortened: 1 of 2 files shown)


robertSt7:
Hi @NiklasBr could you please fix the conflicts? Thanks

templates/admin/login/login.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends '@PimcoreAdmin/admin/login/layout.html.twig' %}
 
 {% block content %}
-<div id="loginform">
+<div id="loginform" {% if not browserSupported %}style="display:none;"{% endif %}>
```
robertSt7, line 4: @NiklasBr Why do you move the css to the inline style?

NiklasBr: image

robertSt7: @NiklasBr What is your opinion about adding this here? https://github.com/pimcore/admin-ui-classic-bundle/pull/227/files#diff-55e6dc29c563c3b12119df18461f947de2b3c73c6dae2cd25a0a2400f0e64ca6L52

NiklasBr: I think it spreads out the logic in an unnecessary way, but that is not a strongly held opinion.

robertSt7: We have the JS part also there, do you see any reason why this shouldn't be together?

NiklasBr: Because this PR was meant to only fix validation errors and not become an everythingburger :)

robertSt7: I totally agree with you, but I don't get the difference of change between moving it to the inline styling or moving it to the `below_footer` block. So you basically need to move it with both approaches, right?

blankse: @robertSt7 `below_footer` is also inside of the body tag. You would need a new block for the head section.

robertSt7: @blankse You are right. Thanks 
@NiklasBr Then we would consider this in a follow up PR

robertSt7:
@NiklasBr Thanks for the PR :)

PR #232, opened 2023-08-24
Content-Security-Policy error on login page

### LINKED ISSUE #228: Content-Security-Policy error on login page
Found in https://github.com/pimcore/pimcore/issues/15781

1. Change User Agent in Firefox (currently v116.0.3) to something unsupported such as the latest version of Opera och Brave.
2. Reload the login page.

Results are the following errors: 
```
Content-Security-Policy: Ignoring “'unsafe-inline'” within script-src: nonce-source or hash-source specified
```
```
Content-Security-Policy: The page's settings blocked the loading of a resource at data:text/css,%5Bid*%3D'google_ads_ifram… ("style-src").
```
```
Content-Security-Policy: The page's settings blocked the loading of a resource at inline ("script-src").
Source: showLogin();
```
```
Content-Security-Policy: The page's settings blocked the loading of a resource at inline ("script-src").
```

robertSt7:
resolves #228

### DIFF
--- a/templates/admin/login/login.html.twig
+++ b/templates/admin/login/login.html.twig
@@ -41,8 +41,6 @@
     <a id="proceed_with_unsupported_browser" href="#" >{{ 'Click here to proceed'|trans([], 'admin') }}</a>
 </div>
 
-<script src="/bundles/pimcoreadmin/js/pimcore/browserNotSupported.js"></script>
-
 <style type="text/css">
     #loginform {
         display: none;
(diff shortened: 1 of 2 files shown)


templates/admin/login/login.html.twig
```diff
@@ -38,15 +38,11 @@
...
+    <a id="proceed_with_unsupported_browser" href="#" >{{ 'Click here to proceed'|trans([], 'admin') }}</a>
 </div>
 
-<script type="text/javascript">
-    function showLogin() {
-        document.getElementById('loginform').style.display = 'block';
-        document.getElementById('browserinfo').style.display = 'none';
-    }
-</script>
+<script src="/bundles/pimcoreadmin/js/pimcore/browserNotSupported.js"></script>
```
kingjia90: LGTM and works fine. 

Just one input: i'd suggest to delete `browserNotSupported.js` and move its content some line below

https://github.com/pimcore/admin-ui-classic-bundle/blob/f97c7f5475c4504cbe1b3793b8ee2af9c600e860/templates/admin/login/login.html.twig#L68

```
{% if not browserSupported %}
document.getElementById('proceed_with_unsupported_browser').addEventListener('click', showLogin);

/**
 * @private
 */
function showLogin() {
    document.getElementById('loginform').style.display = 'block';
    document.getElementById('browserinfo').style.display = 'none';
}
{% endif %}
```

reason being this code is used only here and only when the button `a#proceed_with_unsupported_browser` is present, although won't hurt, but doesn't need to be in an external js as it's never reusable.

kingjia90: Alternatively should move `<script src="/bundles/pimcoreadmin/js/pimcore/browserNotSupported.js"></script>` within `{% if not browserSupported %}` block but not the best practice in terms that the js is neither in header or footer but in middle of the code

NiklasBr: Maybe use https://pimcore.com/docs/platform/Pimcore/MVC/Template/Template_Extensions/HeadScript ?

PR #236, opened 2023-08-24
Fix getDefaultKeyBindings when no user is passed

### LINKED ISSUE #235: [Bug]: Reset key bindings
### Pimcore version

Any

### Steps to reproduce

1. Login as admin
2. Navigate to System settings -> User/Roles -> Users
4. Click on a user
5. Switch to the Key Bindings tab
5. Click the Reset button
Screenshot 2023-08-24 at 15 36 22

### Actual Behavior

An error is thrown when the Reset button is clicked

### Expected Behavior

The Key Bindings are reset to the default state

dvesh3:
Resolves #235

### DIFF
--- a/src/Helper/User.php
+++ b/src/Helper/User.php
@@ -31,7 +31,7 @@ final class User
      */
     public static function getDefaultKeyBindings(Pimcore\Model\User|UserProxy|null $user = null): string
     {
-        if ($user && method_exists($user, 'getKeyBindings') && $user->getKeyBindings()) {
+        if ($user instanceof Pimcore\Model\User && $user->getKeyBindings()) {
             return $user->getKeyBindings();
         }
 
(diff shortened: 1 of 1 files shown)


src/Helper/User.php
```diff
@@ -31,7 +31,7 @@ final class User
      */
     public static function getDefaultKeyBindings(Pimcore\Model\User|UserProxy|null $user = null): string
     {
-        if(method_exists($user, 'getKeyBindings') && $user->getKeyBindings()) {
+        if ($user && method_exists($user, 'getKeyBindings') && $user->getKeyBindings()) {
```
blankse:
```suggestion
        if ($user instanceof Pimcore\Model\User && $user->getKeyBindings()) {
```

blankse:
Target branch should be `1.1`

blankse:
There also are many missing translations:
image

PR #238, opened 2023-08-24
Clean test environment

blankse:
- gitignore should ignore test files (and shouldn't have legacy stuff)
- Composer should set the lowest possible version
- Composer should have the newest dev dependency versions
- .env File is not supported anymore (symfony/dotenv isn't loaded)


dvesh3:
@blankse thanks!

PR #240, opened 2023-08-24
Fix issue with setting the read target to settings-store for Appearance and branding

aryaantony92:
Related to https://github.com/pimcore/pimcore/issues/15596

### DIFF
--- a/src/System/AdminConfig.php
+++ b/src/System/AdminConfig.php
@@ -57,7 +57,18 @@ final class AdminConfig
     {
         $repository = self::getRepository();
 
-        return SystemConfig::getConfigDataByKey($repository, self::CONFIG_ID);
+        $data = SystemConfig::getConfigDataByKey($repository, self::CONFIG_ID);
+        $loadType = $repository->getReadTargets()[0] ?? null;
+
+        // If the read target is settings-store and no data is found there,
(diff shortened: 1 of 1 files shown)


src/System/AdminConfig.php
```diff
@@ -57,7 +57,18 @@ public static function get(): array
...
         $repository = self::getRepository();
 
-        return SystemConfig::getConfigDataByKey($repository, self::CONFIG_ID);
+        $data = SystemConfig::getConfigDataByKey($repository, self::CONFIG_ID);
+        $loadType = $repository->getReadTargets()[0] ?? null;
+
+        // If the read target is settings-store and no data is found there,
+        // load the data from the container config
+        if(!$data && $loadType === $repository::LOCATION_SETTINGS_STORE) {
+            $containerConfig = \Pimcore::getContainer()->getParameter('pimcore_admin.config');
```
dvesh3, line 66: can we get this data from `$repository->getDataFromContainerConfig(self::CONFIG_ID)`?

aryaantony92: No, because this method 'getDataFromContainerConfig()' has private visibility

dvesh3: Got it. thanks!

PR #241, opened 2023-08-25
Check if parent is not undefined before appending it as request param when uploading an asset

siwa-ddiesenreither:
https://github.com/aashan10/pimcore/commit/4d7ed8c1622d1946fea00d3a68061c0cfd525379 this feature introduced a bug. It creates an folder called 'undefined' when no uploadPath is defined in the config object of pimcore_image (inline upload)


siwa-ddiesenreither:
recheck

siwa-ddiesenreither:

robertSt7:
@siwa-ddiesenreither Thanks for the PR

PR #246, opened 2023-08-30
[Bug] Remove the question mark from the Polish translation

wkania:
Other translations, including [English](https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/translations/admin.en.yaml#L102C1-L102C3), do not have a question mark at the end of a sentence. As a result, the Polish translation is displayed with a double question mark.

### DIFF
--- a/translations/admin.pl.yaml
+++ b/translations/admin.pl.yaml
@@ -104,7 +104,7 @@ retype_password: "Potw\xF3rz has\u0142o"
 reports: Raporty
 roles: Grupy
 Password: "Has\u0142o"
-Forgot your password: "Zapomnia\u0142e\u015B has\u0142o?"
+Forgot your password: "Zapomnia\u0142e\u015B has\u0142o"
 Back to Login: "Powr\xF3t do strony logowania"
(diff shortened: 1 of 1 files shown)


blankse:
I think the question mark should be in the translation. Some languages (e. g. greek, arabic) have other question marks.

blankse:
@wkania I created a alternative PR with my suggestion https://github.com/pimcore/admin-ui-classic-bundle/pull/247

wkania:
@blankse I prefer yours PR, because question mark should be part of translation. But now all other translations need to be fixed, because they are missing question mark; It was added after translation in the template.

blankse:
@wkania There are more languages that have already a question mark. Not only Polish. Yes we should add it to all languages if my PR is merged. This is handled via POEditor.

wkania:

wkania:
@blankse Any solution works for me, as long as it is consistent for all languages.

robertSt7:
Closed in favor of https://github.com/pimcore/admin-ui-classic-bundle/pull/247

PR #247, opened 2023-08-30
Question mark of "Forgot your password" should be in the translation

blankse:
See https://github.com/pimcore/admin-ui-classic-bundle/pull/246


robertSt7:
@blankse Thanks :)

PR #249, opened 2023-08-31
Don't use legacy/internal Image::getThumbnailConfig() method

blankse:
See https://github.com/pimcore/pimcore/pull/15869

### DIFF
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -1216,13 +1216,13 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         $thumbnailConfig = null;
 
         if ($request->get('thumbnail')) {
-            $thumbnailConfig = $image->getThumbnailConfig($request->get('thumbnail'));
+            $thumbnailConfig = $image->getThumbnail($request->get('thumbnail'))->getConfig();
         }
         if (!$thumbnailConfig) {
             if ($request->get('config')) {
-                $thumbnailConfig = $image->getThumbnailConfig($this->decodeJson($request->get('config')));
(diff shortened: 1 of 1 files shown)


dvesh3:
@blankse please rebase on 1.x branch as 1.1 requires pimcore/pimcore:^11.0

blankse:
@dvesh3 Why? I think we should make this change ASAP. When we remove the `getThumbnailConfig()` method in pimcore/pimcore, the admin-ui-classic-bundle versions before this change doesn't work anymore. We can make a "conflict" in composer.json in pimcore/pimcore than for this versions.

kingjia90:
this PR should be safe and fine on 1.1.3 as it uses the recommended approach for something internal and deprecated since 6.9, i'd intervene on 11.1 side instead. 
The only problem is admin ui classic 1.0.0 to 1.1.2 won't work with 11.1 if we merge https://github.com/pimcore/pimcore/pull/15869 ,right?
Could the actual removal be moved to 12?

blankse:
@kingjia90 Could we not add this to the composer.json in Pimcore 11.1?
```
  "conflict": {
    "pimcore/admin-ui-classic-bundle": "<1.1.3"
  },
```

Edit: I added this to the PR in Pimcore

kingjia90:
> @kingjia90 Could we not add this to the composer.json in Pimcore 11.1?
> 
> ```
>   "conflict": {
>     "pimcore/admin-ui-classic-bundle": "<1.1.3"
>   },
> ```
> 
> Edit: I added this to the PR in Pimcore

Technically is fine, but personally would use 'conflict' as last resort and it doesn't seem worth it for a clean-up (removing deprecated and "harmelss" method) that could be postponed to next major, would rather consider it for cases where we have to refactor due a severe bug or so.

blankse:
@kingjia90 Ok, than I add a note to remove it in Pimcore 12.

PR #252, opened 2023-08-31
Default value input fields for Multiselect

jan888adams:
Add input fields for default value and default value generator for Multiselect Object type. 

I already have an [MR](https://github.com/pimcore/pimcore/pull/15871) open for the backend part.


public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -99,13 +103,148 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+        this.datax.options = options;
+
+        var defaultValue = [];
+
+        var defaultValueEditor = this.specificPanel.getComponent("defaultvalueeditor");
+        if (defaultValueEditor) {
+            var defaultValueStore = defaultValueEditor.getStore();
+            defaultValueStore.commitChanges();
+            defaultValueStore.each(function (rec) {
+                console.log(rec.get("value"));
```
jdreesen: Leftover debug statement
```suggestion
```

robertSt7:
Hi @jan888adams could you please use `let` and `const` instead of `var` for variables in JavaScript? Thanks

robertSt7:
Hi @jan888adams I get the following errors, when I start pimcore and try to open a class which contains a Select and Multiselect field. 
image
Could you please take a look at this? Thx :)

mattamon:
> Hi @jan888adams I get the following errors, when I start pimcore and try to open a class which contains a Select and Multiselect field. image Could you please take a look at this? Thx :)

I merged the latest 1.x changes into the branch and fixed the class.js error.
Should be good to go now for another review.

mattamon:
@jan888adams can you please recheck this PR. When I add a new multiselect to a class I get the following error:
image

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -99,13 +103,147 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+            },
+            {
+                xtype: "textfield",
+                fieldLabel: t("options_provider_data"),
+                width: 600,
+                value: datax.optionsProviderData,
+                name: "optionsProviderData"
+            },
+            {
+                xtype: "textfield",
```
kingjia90:
```suggestion
                xtype: "hidden",
```
to avoid to show 
image

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -256,84 +394,128 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+                    }
+                })
+            ]
+        });
     },
-
-    showoptioneditor: function (valueStore) {
-        var editor = new pimcore.object.helpers.optionEditor(valueStore);
-        editor.edit();
-    }
```
kingjia90: need to revert this and rename to  `showgrideditor` from `showoptioneditor`, otherwise i get 
image
from https://github.com/pimcore/admin-ui-classic-bundle/pull/252/files#diff-3cabd22d0fd1dda6a0151f733e10f8b4472af24bb72e5c8b8d67dc55a6284335R433-R436

as [pointed out](https://github.com/pimcore/admin-ui-classic-bundle/pull/252#issuecomment-2337428052) by my former coworker

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -256,84 +394,128 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+                        },
+                    }
+                })
+            ]
+        });
     },
-
-    showoptioneditor: function (valueStore) {
-        var editor = new pimcore.object.helpers.optionEditor(valueStore);
-        editor.edit();
```
kingjia90, line 334: might need some tweak because we only have one column in `Default Value`
image
opposed to `Selection Options`
image and the values are all starting with an `,something`

public/js/pimcore/object/classes/data/multiselect.js
kingjia90: or keep it deleted and remove 
https://github.com/pimcore/admin-ui-classic-bundle/pull/252/files#diff-3cabd22d0fd1dda6a0151f733e10f8b4472af24bb72e5c8b8d67dc55a6284335R433-R435
it we don't want to see the comma separated list with `,something` like the other comment

kingjia90 (dismissed):
Thank you for your PR and patience, finally getting close to merge this 😏 

Just  found some minor inputs to be addressed first, other than needing to be rebased and fix the conflict

kingjia90:
An entry about the improvement on https://github.com/pimcore/admin-ui-classic-bundle/blob/2.x/CHANGELOG.md would be also highly appreciated

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -99,13 +103,147 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
-        const valueStore = new Ext.data.JsonStore({
+        const optionsStore = new Ext.data.JsonStore({
             fields: ["key", {name: "value", allowBlank: false}],
             data: datax.options
         });
+        let optionsGrid = this.createoptionsgrid(optionsStore);
+        this.optionsModel = optionsGrid.getSelectionModel();
+
+        const defaultValueStore = new Ext.data.JsonStore({
+            fields: ["value"],
```
kingjia90, line 111: wouldn't be better to be an extra column with a checkbox for every options (`field` in `optionsStore`) to avoid setting up default values that do not match any of the allowed/existing options, instead of own grid?

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -99,13 +103,147 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+        });
+        let defaultValueGrid = this.createdefaultvaluegrid(defaultValueStore);
+        this.defaultValueModel = defaultValueGrid.getSelectionModel();
+
+        let appendedStylingItems = stylingItems.concat([
+            {
+                xtype: "numberfield",
+                fieldLabel: t("maximum_items"),
+                name: "maxItems",
+                value: datax.maxItems,
```
kingjia90, line 122: this can be bypassed by the default value section, so if the limit is 2 and the default are more than 2, eg.4, it would pre-select 4.
From UI/UXpov, it may be better to limit the list of `createdefaultvaluegrid` according to this max value, wdyt?

kingjia90: [Screencast from 2025-06-04 15-33-07.webm](https://github.com/user-attachments/assets/134d32cc-9126-4d05-83bc-558e9674f5bd)

If one then select something, it ends up stuck in a forever loop, popup reappears even by pressing `Ok` on the alert

public/js/pimcore/object/classes/data/multiselect.js
kingjia90: ah nevermind, i see why, it wouldn't work with the other dynamic Options Source type, like Options provider.

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -99,13 +103,147 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
+                    ['tags', 'Tags']
+                ],
+                value: datax["renderType"] ? datax["renderType"] : 'list',
+                triggerAction: "all",
+                editable: false,
+                forceSelection: true
+            },
+            {
+                xtype: "textfield",
+                fieldLabel: t("options_provider_class"),
```
kingjia90, line 142: it seems not working with the other dynamic option provider, can you also check on this?

public/js/pimcore/object/classes/data/multiselect.js
kingjia90: resolved  by https://github.com/pimcore/admin-ui-classic-bundle/pull/252/commits/eba95803ddbbe5f8dcd9476f1393a891dd8009c5#diff-3cabd22d0fd1dda6a0151f733e10f8b4472af24bb72e5c8b8d67dc55a6284335R521

public/js/pimcore/object/classes/data/multiselect.js
kingjia90: resolved by https://github.com/pimcore/admin-ui-classic-bundle/pull/252/commits/1433649aeaeeaf837c8cd3975bc68d20fc1861eb

public/js/pimcore/object/classes/data/multiselect.js
kingjia90: disabling for these type, it's not supported, resolved  by https://github.com/pimcore/admin-ui-classic-bundle/pull/252/commits/c7136ab958127f2a25dfd46dd7b7fc68145f31e5

public/js/pimcore/object/classes/data/multiselect.js
```diff
@@ -253,84 +391,141 @@ pimcore.object.classes.data.multiselect = Class.create(pimcore.object.classes.da
...
-            var valueStore = valueEditor.getStore();
-            valueStore.commitChanges();
-            valueStore.each(function (rec) {
-                options.push({
-                    key: rec.get("key"),
-                    value: rec.get("value")
-                });
-            });
-        }
+                    if (count >= this.datax.maxItems){
```
kingjia90:
```suggestion
                    if (this.datax.maxItems > 0 && count >= this.datax.maxItems){
```
otherwise it would be always reaching the limit

robertSt7 (approved):

PR #255, opened 2023-09-04
Fix: Converting page to link and vice versa

blankse:
Fix https://github.com/pimcore/admin-ui-classic-bundle/issues/258


robertSt7:
@blankse Thanks :)

PR #260, opened 2023-09-06
Fix deleting translation with "%s" character

dvesh3:
### Steps to reproduce

1. Open “Translations” in admin.
2. Add the key “Did you mean: %suggest%?”.
3. Click the “delete” icon.

## Changes
resolves issue with sprintf complaining about not providing enough replacements


kingjia90:
Closing in favor of #264

PR #262, opened 2023-09-07
Fix quicksearch: pimcore_user() only accept ints

blankse:
Alternative to https://github.com/pimcore/pimcore/pull/15914

Related https://github.com/pimcore/pimcore/pull/15915

### DIFF
--- a/templates/searchadmin/search/quicksearch/info_table.html.twig
+++ b/templates/searchadmin/search/quicksearch/info_table.html.twig
@@ -50,7 +50,7 @@
         {% endif %}
 
         {% set userOwnerId = element.getUserOwner() %}
-        {% set owner = userOwnerId ? pimcore_user(userOwnerId) : null %}
+        {% set owner = userOwnerId is not null ? pimcore_user(userOwnerId) : null %}
         {% if owner is instanceof('\\Pimcore\\Model\\User') %}
             <tr>
                 <th>{{ 'owner'|trans([],'admin') }}</th>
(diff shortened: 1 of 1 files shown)


templates/searchadmin/search/quicksearch/info_table.html.twig
```diff
@@ -49,15 +49,17 @@
             {% endif %}
         {% endif %}
 
-        {% set owner = pimcore_user(element.getUserOwner()) %}
+        {% set userOwnerId = element.getUserOwner() %}
+        {% set owner = userOwnerId ? pimcore_user(userOwnerId) : null %}
         {% if owner is instanceof('\\Pimcore\\Model\\User') %}
             <tr>
                 <th>{{ 'owner'|trans([],'admin') }}</th>
```
kingjia90, line 56: this looks like it would break the table column on the rows where the userOwner is missing

templates/searchadmin/search/quicksearch/info_table.html.twig
```diff
@@ -49,15 +49,17 @@
             {% endif %}
         {% endif %}
 
-        {% set owner = pimcore_user(element.getUserOwner()) %}
+        {% set userOwnerId = element.getUserOwner() %}
+        {% set owner = userOwnerId ? pimcore_user(userOwnerId) : null %}
```
kingjia90: this looks like it would break the table column on the rows where the `userOwner` is missing, and `owner` is `null`

blankse: What do you mean? If the owner is null the row is skipped. This was the behaviour also before Pimcore 11.

kingjia90: ahh, you're right.
I've overlooked the `<tr>`, thought it was building a table with multiple columns

bramstroker:
LGTM

PR #263, opened 2023-09-07
Clear controller/template if you convert page to email/snippet

blankse:
See https://github.com/pimcore/pimcore/pull/15910


robertSt7:
@blankse Thanks :)

PR #266, opened 2023-09-08
Fix object rename not updating the tree store

### LINKED ISSUE #203: [Bug] Event pimcore.events.prepareAffectedNodes details not correctly built after renaming assets or data objects

dvesh3:
Resolves #203

### DIFF
--- a/src/Controller/Admin/Document/DocumentController.php
+++ b/src/Controller/Admin/Document/DocumentController.php
@@ -48,6 +48,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpKernel\Event\ControllerEvent;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 /**
--- a/src/Controller/Admin/Document/DocumentController.php
+++ b/src/Controller/Admin/Document/DocumentController.php
@@ -541,8 +541,8 @@ class DocumentController extends ElementControllerBase implements KernelControll
                     return $this->adminJson(['success' => false, 'message' => $e->getMessage()]);
                 }
             } else {
-                $msg = 'prevented moving document, because document with same path+key already exists';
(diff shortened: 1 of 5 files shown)


robertSt7:
@dvesh3 We have the same problem with assets.

dvesh3:
> @dvesh3 We have the same problem with assets.

@robertSt7 fixed it for assets and documents tree. Also reverted the approach for reloading parent on document tree rename.

src/Controller/Admin/Document/DocumentController.php
```diff
@@ -522,15 +530,19 @@ public function updateAction(Request $request): JsonResponse
...
                     if ($oldPath && $oldPath != $document->getRealFullPath()) {
                         $this->firePostMoveEvent($document, $oldDocument, $oldPath);
                     }
                 } catch (Exception $e) {
                     return $this->adminJson(['success' => false, 'message' => $e->getMessage()]);
                 }
             } else {
-                $msg = 'Prevented moving document, because document with same path+key already exists or the document is locked. ID: ' . $document->getId();
+                $msg = 'prevented moving document, because document with same path+key already exists';
+                $msg .= ' aor the document is locked. ID: ' . $document->getId();
```
robertSt7:
```suggestion
                $msg = 'prevented moving document, because document with same path+key already exists' .
                    ' or the document is locked. ID: ' . $document->getId();
```

robertSt7 (changes requested):
@divesh For me the error still occurs with Assets and Documents. Could you please take a second look? Thanks

dvesh3:
> For me the error still occurs with Assets and Documents. Could you please take a second look? Thanks

@robertSt7 yes, there were extra checks in editCompleteKey to check if the element is open before performing reopenelement and these checks are already performed in the called method so removed it.

Note: better compare differences with `/files?diff=split&w=1`

PR #267, opened 2023-09-08
[AdminStyle] Allow element text modification on tree

dvesh3:
Related to https://github.com/pimcore/pimcore/pull/15917

### DIFF
--- a/src/Controller/Traits/AdminStyleTrait.php
+++ b/src/Controller/Traits/AdminStyleTrait.php
@@ -52,8 +52,6 @@ trait AdminStyleTrait
             $data['cls'] .= $adminStyle->getElementCssClass() . ' ';
         }
         $data['qtipCfg'] = $adminStyle->getElementQtipConfig();
-        if (method_exists($adminStyle, 'getElementText')) {
-            $data['text'] = $adminStyle->getElementText();
-        }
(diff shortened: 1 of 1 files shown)


src/Controller/Traits/AdminStyleTrait.php
```diff
@@ -52,5 +52,8 @@ protected function addAdminStyle(ElementInterface $element, int $context = null,
             $data['cls'] .= $adminStyle->getElementCssClass() . ' ';
         }
         $data['qtipCfg'] = $adminStyle->getElementQtipConfig();
+        if (method_exists($adminStyle, 'getElementText')) {
```
robertSt7: Would it be better to use an Interface with get- and setElementText and check then against the Interface instead of `method_exist`?

dvesh3: I was expecting that :laughing:..It was just there to make the tests pass. If decide to bump pimcore/pimcore requirement to 11.1 then we can kick it.

robertSt7: I think then it's ok :) What do you think about adding a short comment, that we can remove this if 11.1 is required?

robertSt7:
waiting for pimcore 11.1 release

Corepex (approved):

PR #269, opened 2023-09-13
[Improvement]: Make Password Recovery Token to be one time use only

kingjia90:
Related and Requires https://github.com/pimcore/pimcore/pull/15942

To be reconsider once the bundle minimum requirement of `pimcore/pimore` is `11.1`, to use a slightly better function.
`generateToken` needs `User` to work (to save the token to the given user) and currently it does a `User::getByName`, but it could be passed straightforwardly like in the PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+#### v1.2.0
+ - Bumped `pimcore/pimcore` minimum requirement to `^11.1.0`
+
 #### v1.1.0
  - `Pimcore\Bundle\AdminBundle\Service\ElementService` is marked as internal.
  - Deprecated `DocumentTreeConfigTrait`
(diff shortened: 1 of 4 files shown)


blankse:
@kingjia90 Is it possible to install admin-ui-classic-bundle `1.0`/`1.1` with the change of pimcore 11.1? I think not, right? So we need to add a conflict to this versions in the composer.json of the core. Because this version have `^11.0` and can install pimcore `11.1.0`.

kingjia90:
> @kingjia90 Is it possible to install admin-ui-classic-bundle `1.0`/`1.1` with the change of pimcore 11.1? I think not, right? So we need to add a conflict to this versions in the composer.json of the core. Because this version have `^11.0` and can install pimcore `11.1.0`.

nice point, true, then it looks would be easier to introduce a new method in 11.1 and "recommending" to use that on here

PR #270, opened 2023-09-14
After renamaning the object in tree it is not getting refreshed.

atul-emico:
When any object in tree is renamed after that that object is not getting refreshed. Because of that if you try to rename that object again then in rename pop up older value is visible not the updated one.

Example - 

Object initial name - test
Rename it to - test one (Object is not getting refreshed)
Again try to rename same object without page or tree refresh, it will show initial value (test) in rename form


atul-emico:

atul-emico:
recheck

dvesh3:
@atul-emico already fixed by https://github.com/pimcore/admin-ui-classic-bundle/pull/266 thanks!

PR #271, opened 2023-09-14
Add missing `AdminEvents::ASSET_GET_FIELD_GRID_CONFIG` dispatch

aryaantony92:


src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -340,6 +339,18 @@ protected function getFieldGridConfig(array $field, string $language = '', strin
...
             $result['layout']['subtype'] = $type;
         }
+        if(!$predefined && $type != 'system') {
+            $eventDispatcher = \Pimcore::getKernel()->getContainer()->get('event_dispatcher');
+            $assetGetFieldGridConfig = new GenericEvent($this, [
+                'field' => $field,
+                'result' => $result
+            ]);
+
+            $eventDispatcher->dispatch($assetGetFieldGridConfig, AdminEvents::ASSET_GET_FIELD_GRID_CONFIG);
```
dvesh3: why dispatch only for non-predefined & non-system?

aryaantony92: To make sure we are only dispatching it for those which belongs to asset meta data class definition

dvesh3: It should be generic so as can be used anywhere not just asset-metadata

aryaantony92: Done in [8f5dc16](https://github.com/pimcore/admin-ui-classic-bundle/pull/271/commits/8f5dc165273d808b1a6dbbcdb5aa631a5eeae897)

src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -340,7 +339,16 @@ protected function getFieldGridConfig(array $field, string $language = '', strin
             $result['layout']['fieldtype'] = 'manyToOneRelation';
             $result['layout']['subtype'] = $type;
         }
-
+        $eventDispatcher = \Pimcore::getKernel()->getContainer()->get('event_dispatcher');
```
dvesh3: Inject the dispatcher instead of getting it directly from the containier:
```
public function __construct(
        protected EventDispatcherInterface $eventDispatcher,
    ) {
    }
```

src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -340,7 +339,16 @@ protected function getFieldGridConfig(array $field, string $language = '', strin
...
+        $assetGetFieldGridConfig = new GenericEvent($this, [
+            'field' => $field,
+            'result' => $result
+        ]);
+
+        $eventDispatcher->dispatch($assetGetFieldGridConfig, AdminEvents::ASSET_GET_FIELD_GRID_CONFIG);
+        if(isset($assetGetFieldGridConfig['fieldData'])) {
```
dvesh3: why not pass layout and set it back from the event?

dvesh3: Oh, we already pass result so should be reset from that argument.

PR #273, opened 2023-09-14
[Bug]: Fix versions reload should not trigger on failed save

### LINKED ISSUE #272: [Improvement]: Avoid version reload if saving failed
### Improvement description

followup: pimcore/pimcore#13609

If you save a data object that got deleted, you get two error popups.
Maybe we can avoid the version reload if saving returns `status: false`.

image

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/272

Not sure if it has behaviour change implications, but looks buggy, i don't understand what's the[ try catch block ](https://github.com/pimcore/admin-ui-classic-bundle/pull/273/files?w=1#diff-19561231ee9a8f94a4e06afe4d726dd9f4be310a6ab04e3ed22d9289e9f1fc45L856)is doing, there's a long-time TODO there

forceReloadVersionsAfterSave is for manual overriding via event `pimcore.events.postSaveObject` by setting `e.detail.object.forceReloadVersionsAfterSave = true;` in the listener func

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+#### v1.2.0
+ - DataObject used to automatically reloaded after save in any circumstances, now it's triggered only on successfull save. The reload can be forced by setting `forceReloadAfterSave` to `true` in a `postSaveObject` event listener.
+
 #### v1.1.0
  - `Pimcore\Bundle\AdminBundle\Service\ElementService` is marked as internal.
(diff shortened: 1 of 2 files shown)


robertSt7:
@kingjia90 Could you please rebase it to 1.x and what do you think about adding some docs?

kingjia90:
@robertSt7 Done! Feel free to change/apply directly if i missed something

PR #274, opened 2023-09-15
Use Maintenance Mode Helper

dvesh3:
Related to https://github.com/pimcore/pimcore/pull/15954

### DIFF
--- a/src/Controller/Admin/MiscController.php
+++ b/src/Controller/Admin/MiscController.php
@@ -262,6 +262,9 @@ class MiscController extends AdminAbstractController
         }
 
         if ($request->get('deactivate')) {
+            if (Tool\Admin::isInMaintenanceMode()){
+               Tool\Admin::deactivateMaintenanceMode();
+            }
             $maintenanceModeHelper->deactivate();
         }
 
--- a/src/Controller/Admin/MiscController.php
+++ b/src/Controller/Admin/MiscController.php
@@ -253,16 +253,16 @@ class MiscController extends AdminAbstractController
      *
      * @return JsonResponse
      */
-    public function maintenanceAction(Request $request): JsonResponse
+    public function maintenanceAction(Request $request, Tool\MaintenanceModeHelperInterface $maintenanceModeHelper): JsonResponse
     {
         $this->checkPermission('maintenance_mode');
 
         if ($request->get('activate')) {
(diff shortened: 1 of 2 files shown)


src/Controller/Admin/MiscController.php
```diff
@@ -253,16 +253,16 @@ public function getValidFilenameAction(Request $request): JsonResponse
...
         $this->checkPermission('maintenance_mode');
 
         if ($request->get('activate')) {
-            Tool\Admin::activateMaintenanceMode($request->getSession()->getId());
+            $maintenanceModeHelper->activate($request->getSession()->getId());
         }
 
         if ($request->get('deactivate')) {
-            Tool\Admin::deactivateMaintenanceMode();
+            $maintenanceModeHelper->deactivate();
```
kingjia90, line 268:
```suggestion
            if (Tool\Admin::isInMaintenanceMode()){
               Tool\Admin::deactivateMaintenanceMode();
            }
            $maintenanceModeHelper->deactivate();
```
more edge case but without this, if  you were already in legacy maintenance mode, you would see the icon but it won't exit by clicking it
image
and would break script that may be calling `pimcore.helpers.deactivateMaintenance`.

kingjia90 (changes requested):
LGTM just one particular case that could be covered by deactivating both approaches, until we remove the support for legacy mode and there is missing a followup for this
https://github.com/pimcore/admin-ui-classic-bundle/blob/82aa7a8e6e5a8157c8ff96e027be3a75a9bfd9d2/src/Security/Authenticator/AdminAbstractAuthenticator.php#L106-L109

src/Controller/Admin/MiscController.php
dvesh3: Good catch! yes, we need to check if the maintenance file exists first.

kingjia90 (approved):

dvesh3:
> LGTM just one particular case that could be covered by deactivating both approaches, until we remove the support for legacy mode and there is missing a followup for this
> 
> https://github.com/pimcore/admin-ui-classic-bundle/blob/82aa7a8e6e5a8157c8ff96e027be3a75a9bfd9d2/src/Security/Authenticator/AdminAbstractAuthenticator.php#L106-L109

This is deprecated and will be removed in Pimcore 12. please see https://github.com/pimcore/pimcore/pull/15991

PR #285, opened 2023-09-21
[Improvement]: Check if new password is NOT the same as the old one when resetting

kingjia90:
Password should be different when resetting
https://i.imgflip.com/2frsne.jpg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 #### v1.2.0
  - DataObject used to automatically reload version after save, but now it's triggered only on successfull save. The reload can be forced by setting `forceReloadVersionsAfterSave` to `true` in a `postSaveObject` event listener.
+ - [Auth] When resetting password, setting the new password as the same as the old one would throw an error.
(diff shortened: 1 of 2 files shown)


dvesh3 (changes requested):
@kingjia90 please rebase as #286 is merged and changes merged into 1.x branch. Also, add upgrade notes about the changed behavior in UI.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 #### v1.2.0
  - DataObject used to automatically reload version after save, but now it's triggered only on successfull save. The reload can be forced by setting `forceReloadVersionsAfterSave` to `true` in a `postSaveObject` event listener.
+ - [Auth] When resetting password, setting the new password as the same as the old one would throw an error.
```
dvesh3:
```suggestion
 - [User -> Settings] When resetting password, setting the new password same as the old one would throw an error.
```

dvesh3 (approved):

PR #288, opened 2023-09-21
[Events] add event pimcore.events.postBuildPerspectiveElementTree and…

samyemad:
Resolves  https://pimcore.atlassian.net/browse/PSDP-326

### WHAT
1- Add event pimcore.events.postBuildPerspectiveElementTree
2- Allow another type of element tree in perspective ( alternative_element_tree ) 

### HOW
copilot:walkthrough

### DIFF
--- a/public/js/pimcore/startup.js
+++ b/public/js/pimcore/startup.js
@@ -825,7 +825,6 @@ Ext.onReady(function () {
                     if (!treeConfig.hidden) {
                         customPerspectiveElementTrees.push(treeConfig);
                     }
-                    break;
             }
 
 
--- a/public/js/pimcore/startup.js
+++ b/public/js/pimcore/startup.js
@@ -821,7 +821,7 @@ Ext.onReady(function () {
                         }
                     }
                     break;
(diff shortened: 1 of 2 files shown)


public/js/pimcore/startup.js
```diff
@@ -822,6 +821,10 @@ Ext.onReady(function () {
                         }
                     }
                     break;
+                default:
+                    if (!treeConfig.hidden) {
```
dvesh3, line 825: Should we really check for `hidden` property? or simply add it to the custom trees?

samyemad: @dvesh3 we should check for hidden property before adding it to custom trees 
 if this Perspective element tree is hidden so why i will add it in the custom trees and add it to our event and then check is hidden or not after I listen to this event , and in all cases it will not create tree to this perspective element , so i didn't add it to custom trees array

dvesh3: I think the event listener should decide what to do with hidden trees. It might be required in some cases that you deal with hidden trees and add them to tree panel on some actions.

PR #296, opened 2023-09-26
[Workflows] Fix workflowTransitionHandler not considering global actions

orjansandell:
Global actions get treated as a normal transition since it does not get the attribute isGlobalAction set to true. This causes the global action to be posted to the wrong endpoint with an error saying unknown transition.


orjansandell:

orjansandell:
recheck

orjansandell:
recheck

kingjia90:
Thank you! 

Side notes: Looks a regression from https://github.com/pimcore/pimcore/pull/13018
Deleted CLA recheck comments

PR #302, opened 2023-10-04
[Improvement]: Add the possibility to select type for Calculated Values

SamyMP:

This pull request adds different type for calculated values. 
Sometimes, your calculator returns a boolean or a number, it should be possible to filter the value as a number instead of a string (adds greater than, less than...) in grids

image

### DIFF
--- a/public/js/pimcore/object/classes/data/calculatedValue.js
+++ b/public/js/pimcore/object/classes/data/calculatedValue.js
@@ -115,7 +115,6 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
                         ['html', t('html')],
                         ['number', t('number')],
                         ['date', t('date')],
-                        ['list', t('select')],
                         ['boolean', t('bool')]
                     ]
                 },
--- a/public/js/pimcore/object/classes/data/calculatedValue.js
+++ b/public/js/pimcore/object/classes/data/calculatedValue.js
@@ -112,7 +112,11 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
                     store: [
                         ['input', t('input')],
                         ['textarea', t('textarea')],
-                        ['html', t('html')]
+                        ['html', t('html')],
+                        ['number', t('number')],
+                        ['date', t('date')],
+                        ['list', t('select')],
+                        ['boolean', t('bool')]
                     ]
                 },
                 {
--- a/public/js/pimcore/object/classes/data/calculatedValue.js
+++ b/public/js/pimcore/object/classes/data/calculatedValue.js
@@ -113,7 +113,7 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
                         ['input', t('input')],
                         ['textarea', t('textarea')],
                         ['html', t('html')],
-                        ['number', t('number')],
+                        ['numeric', t('number')],
                         ['date', t('date')],
                         ['boolean', t('bool')]
                     ]
--- a/public/js/pimcore/object/tags/calculatedValue.js
+++ b/public/js/pimcore/object/tags/calculatedValue.js
@@ -107,15 +107,16 @@ pimcore.object.tags.calculatedValue = Class.create(pimcore.object.tags.abstract,
             }
 
             if (value && this.fieldConfig?.elementType === 'date') {
-                const timestamp = intval(value) * 1000;
-                const date = new Date(timestamp);
-
-                return Ext.Date.format(date, "Y-m-d");
+                if (!isNaN(+value)) {
+                    const timestamp = parseInt(value) * 1000;
+                    const date = new Date(timestamp);
+                    return Ext.Date.format(date, "Y-m-d");
+                }
             } else if (this.fieldConfig?.elementType === 'boolean') {
-                if (value) {
-                    return "true"
+                if (this.fieldConfig.calculatorType !== "expression") {
+                    return value ? "true" : "false";
                 } else {
-                    return "false"
+                    return JSON.parse(value) ? "true" : "false";
                 }
             }
             else if (value && (this.fieldConfig === undefined || this.fieldConfig.elementType !== 'html')) {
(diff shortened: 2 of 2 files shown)


fashxp:
should we also (optionally) consider that in getter return types for generated php classes?

SamyMP:
Hello @fashxp, 

I also thought about that. In fact, I event thought it was needed in order to get this working, but in the end, it just needed some javascript (9 lines...)

So, I pushed this as it is for now because it easily permits to choose the filter you want for your calculated values, out of the box, without breaking changes or needing to change anything in your implementation

As a follow-up, we could eventually do what you suggest to make sure your calculated value returns the type you selected, what do you think ?

SamyMP:
I just did some test, we would need to change some methods from :

```php
    public static function getCalculatedFieldValueForEditMode(Concrete $object, array $params, ?Data\CalculatedValue $data): ?string
```

to : 

```php
    public static function getCalculatedFieldValueForEditMode(Concrete $object, array $params, ?Data\CalculatedValue $data): string|int|bool|\DateTime|null
```

Here for example :
https://github.com/pimcore/pimcore/blob/402fdab2a330dfac6800be348707d9c4b897a737/models/DataObject/Service.php#L1446

Is this the way to go ? Or should we juste put "mixed" ?

The `getReturnTypeDeclaration` would also look like this :

```php
public function getReturnTypeDeclaration(): ?string
    {
        if ($this->elementType === 'number') {
            return 'int|float|null';
        } elseif ($this->elementType === 'date') {
            return '?\DateTime';
        } elseif ($this->elementType === 'bool') {
            return '?bool';
        } else {
            return '?string';
        }
    }
```

https://github.com/pimcore/pimcore/blob/402fdab2a330dfac6800be348707d9c4b897a737/models/DataObject/ClassDefinition/Data/CalculatedValue.php#L366C29-L366C29

And there are probably other changes to consider, like the `getPhpdocReturnType` method

fashxp:
Ok, I think it was a bad idea, as the interface [`CalculatorClassInterface`](https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/ClassDefinition/CalculatorClassInterface.php#L24) always returns a string ... so we would need to cast back and forth or change a lot of signatures (that would be a bigger change). 
Let's keep the scope of that PR as it is 👍

robertSt7:
@SamyMP I have tried to filter by the calculated values, but it doesn't work. Could you please take a second look? Thanks

SamyMP:
Hi @robertSt7,

Could you be more precise about what's not working ?

Two things to note : 
First, due to the nature of how calculated values work, it may be possible in some cases that you need to save every objects (for exemple, cars) in order for the filters to work as intended. This is because the calculated value is only displayed when it's calculated, it persists on the database only when an object is saved. For exemple, if I choose Boolean filter for 'attributesAvailable', save the CAR class, modify the QualityCalculator as follow : 

```php
// QualityCalculator.php 
// [...]
if ($context->getFieldname() == 'attributesAvailable') {
    return $object->getAttributes() && $object->getAttributes()->getItems();
}
// [...]
```

And filter on "Yes" or "No" in the grid, this won't work because if you run the following SQL Query on the database : 
```sql
select distinct attributesAvailable from object_CAR;
```
You'll see `completed` and `not completed` instead of seeing `0` and `1`. The calculated value is only persisted on the database when the object is saved, even though it displays something else on the screen.


Second, I have tried the filters and the only problem I found is that the "select" option doesn't work as intended. It only displays the options that are displayed in the grid as filter possibilities. For exemple : 

image

It only displays "completed" because this is the only value displayed on the grid for this field. If I sort the values to also display the "not competed", I can filter on both.

image

Apart from this caveat, the other filters are working as intended. I can try to work on that and see if it's possible to fix the select filter. In the meantime, what we could do is remove this filter to merge this PR anyway since this is the only problem I was able to identify. What are your thoughts on that ?

robertSt7:
Hi @SamyMP In my opinion we should remove the select filter for now, that we can merge this PR. Could you please do that?

SamyMP:
It's done @robertSt7 😉

robertSt7:
@SamyMP When I try to filtering a number type like this. It doesn't work, but I am not sure why. Could you please have a short look? Thanks
image

SamyMP:
Hi @robertSt7,

I made the corresponding modifications. I needed to change one thing in the Pimcore repo (see [this PR](https://github.com/pimcore/pimcore/pull/16144)). 

I think it's ready for review

robertSt7:
Hi @SamyMP thanks the number-filter works now. I have some doubts about the filtering with boolean. Could you please have a another look? Thanks
image

SamyMP:
Hi @robertSt7, it's weird, it's working for me with this setup :

```php
// Quality calculator
if ($context->getFieldname() == 'saleInformationAvailable') {
    if ($object->getSaleInformation() && $object->getSaleInformation()->getSaleInformation()) {
        return true;
     } else {
        return false;
     }
}
```

and in the class: 
image


But the grid displays 1's or empty cells, not true/false

image


It works when I filter though (after I saved some objects like I explained in a previous message, due to how Calculated values work).
I'll try to work on the display (true/false) in the grid

SamyMP:
Hi @robertSt7, could you please try again ? Don't forget to save the DataObjets that use a calculated value of type boolean.

image


image

robertSt7:
Hi @SamyMP thank you for the answer. The filter works with the class, but it doesn't work for me with the Expression. Could you please take a look at this? Thanks
image

SamyMP:
Hi @robertSt7,

I have changed a few things to take into consideration the calculator type while filtering. While it is possible to return boolean instead of a string in `QualityCalculator.php` when you use  `Class`, you have to use strings when you use `Expression`. It's ready to be reviewed

Some changes also happened in the [pimcore/pimcore repo](https://github.com/pimcore/pimcore/pull/16144) (by the way, thank you @NiklasBr for the suggestion earlier on that PR)

robertSt7:
Hi @SamyMP thanks a lot for your effort :)

PR #307, opened 2023-10-11
[Improvement]: Add post edit events

robertSt7:
see https://github.com/pimcore/admin-ui-classic-bundle/issues/203

### DIFF
--- a/public/js/pimcore/events.js
+++ b/public/js/pimcore/events.js
@@ -298,6 +298,18 @@ pimcore.events.globalLanguageChanged = "pimcore.globalLanguage.changed";
 pimcore.events.postEditObjectKey = "pimcore.objectKey.postEdit";
 
 /**
+ * fired when asset key is edited
+ * object and key are passed as parameter
+ */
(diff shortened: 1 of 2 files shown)


public/js/pimcore/events.js
```diff
@@ -297,6 +297,18 @@ pimcore.events.globalLanguageChanged = "pimcore.globalLanguage.changed";
  */
 pimcore.events.postEditObjectKey = "pimcore.objectKey.postEdit";
 
+/**
+ * fired when asset key is edited
+ * object and key are passed as parameter
```
dvesh3:
```suggestion
 * asset and key are passed as parameter
```

public/js/pimcore/events.js
```diff
@@ -297,6 +297,18 @@ pimcore.events.globalLanguageChanged = "pimcore.globalLanguage.changed";
...
+ */
+pimcore.events.postEditAssetKey = "pimcore.assetKey.postEdit";
+
+/**
+ * fired when document key is edited
+ * object and key are passed as parameter
```
dvesh3:
```suggestion
 * document and key are passed as parameter
```

PR #311, opened 2023-10-16
[Bug]: WYSIWYG Editable is not empty when no content is added

### LINKED ISSUE #304: [Bug]: WYSIWYG Editable is not empty when no content is added
After adding some content to a WYSIWYG editable and removing it again it is not cleared correctly. The persisted value is `<p><br /></p>` but it should be empty. This seems to be the same issue as https://github.com/pimcore/admin-ui-classic-bundle/issues/182 just at a different place.

robertSt7:
resolves #304

### DIFF
--- a/public/js/pimcore/document/edit.js
+++ b/public/js/pimcore/document/edit.js
@@ -323,11 +323,20 @@ pimcore.document.edit = Class.create({
         Object.values(this.getEditables()).forEach(editable => {
             try {
                 if (editable.getName() && !editable.getInherited()) {
-                    let name = editable.getName();
-                    values[name] = {
-                        data: editable.getValue(),
-                        type: editable.getType()
-                    };
+                    const name = editable.getName();
+                    const type = editable.getType();
+                    const value = editable.getValue();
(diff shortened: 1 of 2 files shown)


aryaantony92 (approved):

public/js/pimcore/document/edit.js
```diff
@@ -323,11 +323,21 @@ pimcore.document.edit = Class.create({
...
-                    let name = editable.getName();
-                    values[name] = {
-                        data: editable.getValue(),
-                        type: editable.getType()
-                    };
+                    const name = editable.getName();
+                    const type = editable.getType();
+                    const value = editable.getValue();
+                    if (type === "wysiwyg" &&
+                        (value === "<p><br></p>" || value === "<p><br data-mce-bogus=\"1\"></p>")) {
```
brusch, line 330: @robertSt7 wouldn't it be better to use something more generic, like `empty(trim(strip_tags(value))` ?

robertSt7: @brusch Thanks for the suggestion. I will create a PR for that

PR #314, opened 2023-10-17
Focal Point Marker vanishes on resize

jheimbach:
If the window is resized the focal point marker element is removed from the image container ( `Ext.get(this.previewContainerId).setHtml(html);`) but the internal reference to the marker element is kept and the `addFocalPoint` method will only add a new marker if the reference is falsy

## Steps to reproduce
 - open image asset
 - set focal point
 - resize window

### DIFF
--- a/public/js/pimcore/asset/image.js
+++ b/public/js/pimcore/asset/image.js
@@ -379,6 +379,7 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
 
         var html = '<img src="' + this.data.imageInfo['previewUrl'] + '">';
         Ext.get(this.previewContainerId).setHtml(html);
+        this.marker = false;
 
         let area = this.displayPanel.getEl().down('img');
         if(area) {


jheimbach:
**_I have read the CLA Document and I hereby sign the CLA_**

jheimbach:
recheck

blankse:
@jheimbach The target branch should be `1.2` because it is a bug fix.

jheimbach:
Hey @blankse, i changed the base branch

kingjia90:
Thank you! 

Update: will open a new issue/PR for these

The focal point is now fine, but during review, this PR helped discovering other bugs that we didn't notice before.

https://github.com/pimcore/admin-ui-classic-bundle/pull/314/files?w=1#diff-c2c48ecf0d39cbbefc133304ced7a4b5015d869732c595fba8ebc146449ef3c2R390-R394 these lines bring the focal point to the initial point (before the resize) on resize. 
To reproduce, move the focal point to point (Z), save, move the saved focal point somewhere else (Y) and resize the screen, it would bring the focal point to the saved point (Z) instead of recalculating the new relative position of (Y) of the draft. 
Ideally would be fixed by creating a method that extracts the coordinates from `getSaveData` below and remember it before resize and reposition it on `initPreviewImage`/resize

Also, it looks like there is no way to remove focal point once set.

~Do you have the chance to add these changes in this PR? Otherwise, i'll open a new issue/pr for these. Thank you in advance~

PR #320, opened 2023-10-25
Fix: clean-up string

robertSt7:


dvesh3:
@robertSt7 why do we need this fix?

robertSt7:
Closed in favor of https://github.com/pimcore/pimcore/pull/16211

PR #324, opened 2023-10-27
Make two top rows of versions table sticky

NiklasBr:
I just think this would be nice because I keep forgetting which version is on which side.

https://github.com/pimcore/admin-ui-classic-bundle/assets/279826/21af0c08-7522-4bb4-9b71-21a243439fbc

### DIFF
--- a/public/css/object_versions.css
+++ b/public/css/object_versions.css
@@ -26,6 +26,18 @@ body {
 	width: 100%;
 }
 
+.preview tr:nth-of-type(1) {
+    position: sticky;
+    top: 0;
+    height: 24px;
+}
+.preview tr:nth-of-type(2) {
+    position: sticky;
+    top: 24px; /* :nth-of-type(1) height */
+    min-height: 24px;
+    background-color: #f6f6f6;
+}
+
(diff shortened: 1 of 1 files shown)


NiklasBr:
I understand community contributions are no longer of value.

brusch:
Why? We had the review of this PR in our backlog?

kingjia90:
@NiklasBr wait, please, don't "rush to conclusions" yet, there will be an official communication in the beginning of 2024 on how it is going to be in the future (I'm also waiting for it and unfortunately have nothing to share about), but it's not true that community contribution are no longer of value. 

This PR in particular is for next minor release, therefore we didn't have the chance to review or take care of it due other stuff getting higher priority but so far looks quite nice

fashxp:
> I understand community contributions are no longer of value.

Community contributions are still highly value, that has never changed. 
However, we are closing older, inactive issues that we cannot address either, to better manage resources and maintain focus. This approach also helps to clarify our focus topics to the community. 
Regrettably, if contributors close their relevant issues and pull requests, we must accept it. But it won’t help anybody though.

PR #329, opened 2023-10-31
fix wysiwyg editor init var

### LINKED ISSUE #334: [Bug]: Wysiwyg tinyMCE: HTML markup not being saved
### Pimcore version

11

### Steps to reproduce

I encountered a problem in Pimcore 11 concerning the tinyMCE editor. When saving data, it seems the system stores the content without the HTML markup. This behavior is triggered by the keyup event.

Steps to Reproduce:

1. Navigate to the demo at https://demo.pimcore.com/admin/
2. Open any object with a WYSIWYG field, e.g., "Event".
3. Click on the WYSIWYG field and move the cursor either left or right.
4. Autosave will get triggered, sending the data without HTML markup.
5. Refresh the page.
6. The object wysiwyg field doesn't have HTML formatting.

### Actual Behavior

System didn't save the data in HTML format but in pure plain text

### Expected Behavior

The object should save data with HTML format.

piotrecky:
Fixed https://github.com/pimcore/admin-ui-classic-bundle/issues/334

### DIFF
--- a/public/js/pimcore/startup.js
+++ b/public/js/pimcore/startup.js
@@ -971,5 +971,7 @@ pimcore.helpers.unload = function () {
 };
 
 L.Icon.Default.imagePath = '../bundles/pimcoreadmin/build/admin/images/';
-pimcore.wysiwyg = {};
-pimcore.wysiwyg.editors = [];
+if (!pimcore.wysiwyg) {
+    pimcore.wysiwyg = {};
+    pimcore.wysiwyg.editors = [];
+}


piotrecky:

piotrecky:
recheck

piotrecky:
recheck

public/js/pimcore/startup.js
```diff
@@ -971,5 +971,7 @@ pimcore.helpers.unload = function () {
 };
 
 L.Icon.Default.imagePath = '../bundles/pimcoreadmin/build/admin/images/';
-pimcore.wysiwyg = {};
-pimcore.wysiwyg.editors = [];
+if (!pimcore.wysiwyg) {
```
kingjia90, line 974: doesn't look the right fix, it is just disabling this block this https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/pimcore/object/tags/wysiwyg.js#L182-L186

kingjia90: re-assigning to @robertSt7 as he is more familiar with these changes

ps-omegacode: without this, the editor in objects does not see the wysiwyg setting

piotrecky: without this, the editor in objects does not see the wysiwyg setting

kingjia90: ah ok, i see what you mean, then it's valid

kingjia90:
wrong base branch, cherry picked into 1.2 with https://github.com/pimcore/admin-ui-classic-bundle/commit/38b1f2419f20f2af846c6cbcaa04f7db02665d43

PR #331, opened 2023-11-02
[Assets]: Fix focal point position on screen resize and add button to remove focal point

kingjia90:
Follow up to https://github.com/pimcore/admin-ui-classic-bundle/pull/314

Please see https://github.com/pimcore/admin-ui-classic-bundle/pull/314#issuecomment-1785231281


public/js/pimcore/asset/image.js
```diff
@@ -427,30 +445,48 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
...
-    },
+        Ext.getCmp('remove_focal_point').setVisible(true);
+        Ext.getCmp('add_focal_point').setVisible(false);
 
-    getSaveData : function ($super, only) {
-        var parameters = $super(only);
+        this.marker = marker;
 
+    },
+    removeFocalPoint: function () {
```
jheimbach, line 456: in `addFocalPoint`,  a context menu on the focal point element (line 425) is added that has the logic to delete a focal point, you should maybe replace the anonymous handler with this method.

kingjia90: Thank you!
https://github.com/pimcore/admin-ui-classic-bundle/pull/331/commits/15685e1ed732d09dc5bbeb3ff2b017b4e587680c

dvesh3 (changes requested):
Bug: can't open 2 images together
image

public/js/pimcore/asset/image.js
```diff
@@ -427,30 +444,48 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
...
         if(this["marker"]) {
+            this.marker.remove();
+            this["marker"] = false;
+            Ext.getCmp('remove_focal_point_' + this.id).setVisible(false);
+            Ext.getCmp('add_focal_point_' + this.id).setVisible(true);
+        }
+    },
+    getFocalPointCoordinates: function () {
```
dvesh3: maybe change the function name to something not `get` as it is not returning the coordinates.

public/js/pimcore/asset/image.js
```diff
@@ -427,30 +444,48 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
...
 
-            var boundingBox = this.marker.up().getSize();
+            let boundingBox = this.marker.up().getSize();
 
-            var x = round(left * 100 / boundingBox.width, 8);
-            var y = round(top  * 100 / boundingBox.height, 8);
+            x = round(left * 100 / boundingBox.width, 8);
+            y = round(top * 100 / boundingBox.height, 8);
+        }
+        this.focalPointCoordinates = {x: x, y: y};
```
dvesh3: Set this only when recalculation happens as the property is initialized with same default values. then also get rid of x & y defaults.

public/js/pimcore/asset/image.js
kingjia90: thank you for the feedback, changed to set on https://github.com/pimcore/admin-ui-classic-bundle/pull/331/commits/b95886e193b0f31069fbe7973d190047adf88374

kingjia90: ~oh wait, maybe something like "calculateFocalPointCoordinates" because it's not a setter (that passes values) either.~ ended up with `updateFocalPointCoordinates`

dvesh3 (changes requested):
Another bug:
1. Set Focal point and save & reload asset.
3. Click Remove focal point and save Asset.
4. Resize window, the focal point appears again.

dvesh3 (approved):

dvesh3:
thanks @kingjia90 :wink:

PR #333, opened 2023-11-02
[Bug]: Fixes DataObject global language selector not working

### LINKED ISSUE #332: [Bug]: Toolbar Language Selector
### Pimcore version

11

### Steps to reproduce

When you select a language in the toolbar the content changes in the object  localized tab with the fields from the selected language, but when you add more than 15 languages, those tabs change to a select, and when you do the same process, the localized content doesn't switch.

### Actual Behavior

When you select a language in the toolbar the content changes in the object  localized tab with the fields from the selected language, but when you add more than 15 languages, those tabs change to a select, and when you do the same process, the localized content doesn't switch.

### Expected Behavior

It should do the same when the localized fields in the object are tabs or select. I did a patch to solve this. In the localizedfields.js line 286, pimcoreGlobalLanguageChanged you should add part of the functionality is on the listeners above, because is not entering in that listener when the value of the select changes. So adding this it works:

this.availablePanels[this.currentLanguage].hide();
this.component.updateLayout();
this.availablePanels[language].show();
this.currentLanguage = language;
this.component.updateLayout();

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/332

### DIFF
--- a/public/js/pimcore/object/tags/localizedfields.js
+++ b/public/js/pimcore/object/tags/localizedfields.js
@@ -276,7 +276,7 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
                             if (oldLanguage == newLanguage) {
                                 return;
                             }
-                            this.refreshLocalizedPanels(oldLanguage, newLanguage);
+                            this.switchLocalizedPanels(oldLanguage, newLanguage);
 
                         }.bind(this),
                         pimcoreGlobalLanguageChanged: function (language) {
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/localizedfields.js
```diff
@@ -289,6 +289,12 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
                                 globalLanguageIndex = this.frontendLanguages.indexOf(language);
                             }
                             this.countrySelect.setValue(this.frontendLanguages[globalLanguageIndex]);
+
+                            this.availablePanels[this.currentLanguage].hide();
+                            this.component.updateLayout();
+                            this.availablePanels[language].show();
+                            this.currentLanguage = language;
+                            this.component.updateLayout();
```
dvesh3: maybe extract the logic to a method and call it in `select` and `pimcoreGlobalLanguageChanged` event listeners.

dvesh3 (approved):

PR #335, opened 2023-11-06
Fix: Grid export results in error when inheritance is disabled

### LINKED ISSUE #336: Grid export results in error when inheritance is disabled
### Discussed in https://github.com/pimcore/pimcore/discussions/16200

<div type='discussions-op-text'>

<sup>Originally posted by **khenow1978** November  6, 2023</sup>
Hi.. in Pimcore there's something that bugs me (no pun intended) since our update to v11.0.12. 

If you try to export any grid view to xlsx with disabled inheritance it results in this error:

image

It looks as if the value for the "GetInheritedValue" function isn't populated properly. Unfortunately I'm not a programmer and can't resolve it by myself.. is anybody able to reproduce this error?</div>

blankse:
Fix https://github.com/pimcore/admin-ui-classic-bundle/issues/336

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/src/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1302,7 +1302,7 @@ class DataObjectHelperController extends AdminAbstractController
 
         $allParams = array_merge($request->request->all(), $request->query->all());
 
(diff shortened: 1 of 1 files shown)


robertSt7:
@blankse Thanks

PR #339, opened 2023-11-08
Fix: Use LocaleServiceInterface instead LocaleService

blankse:
Fix: https://github.com/pimcore/pimcore/issues/16215

### DIFF
--- a/src/DataObject/GridColumnConfig/Value/AbstractValue.php
+++ b/src/DataObject/GridColumnConfig/Value/AbstractValue.php
@@ -23,9 +23,9 @@ abstract class AbstractValue implements ValueInterface
 
     protected string $label;
 
-    protected mixed $context = null;
+    protected mixed $context;
 
-    public function __construct(\stdClass $config, mixed $context = null)
+    public function __construct(\stdClass $config, mixed $context)
     {
         $this->attribute = $config->attribute;
         $this->label = $config->label;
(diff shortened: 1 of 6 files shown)


src/DataObject/GridColumnConfig/Value/AbstractValue.php
```diff
@@ -23,9 +23,9 @@ abstract class AbstractValue implements ValueInterface
 
     protected string $label;
 
-    protected mixed $context = null;
+    protected mixed $context;
 
-    public function __construct(\stdClass $config, mixed $context = null)
+    public function __construct(\stdClass $config, mixed $context)
```
robertSt7: @blankse For me this looks like a breaking change. Is there a special reason why do you changed it?

blankse: @robertSt7 Because the $localService param in the DefaultValue class must always be set. See the usage of it. A optional param should not before a required param. 

The params are always set in the factory class. See:
https://github.com/pimcore/admin-ui-classic-bundle/blob/3c7f80d32129837d42a8b198a1e0de5daa210f79/src/DataObject/GridColumnConfig/Value/Factory/DefaultValueFactory.php#L37

Alternative is to create a setLocaleService() method in the DefaultValue, so the constructor can have a optional $context param. Should I change it to this?

robertSt7: @blankse That would be great if you create a setLocaleService() method, that we don't change the `AbstractValue`. Thanks

blankse: @robertSt7 Done: https://github.com/pimcore/admin-ui-classic-bundle/pull/339/commits/e1ae19c5c2e9c9c12ad19cc81ec2eb58c408f324

robertSt7:
@blankse Thanks a lot

PR #341, opened 2023-11-08
[Translation]: Add `username` translation in login form

### LINKED ISSUE #202: Incorrect loading of translations for username and password on login screen
On the login screen, there's a slight difference between the placeholders for `username` and `password`:

image

While this isn't much of an issue in English, in German it shows up as "username" and "Passwort".

There's an admin translation for the key `password` (and also for `Password`...) in `translations/admin.en.yaml` however there's none for `username`; that one only exists in `translations/admin_ext.en.yaml`.

What's the correct fix here, to load/merge translations from `admin_ext.*.*yaml` or to duplicate the relevant key from `admin_ext.*.yaml` to `admin.*.yaml`?

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/202

It would be easier to copy the translation of `Username` (first letter capital) for lowercase username
Altenratively could change the placeholder
https://github.com/pimcore/admin-ui-classic-bundle/blob/bed29a3cc0321c1065a3ab43af73a32a4627f3af/templates/admin/login/login.html.twig#L18-L19

with `{{ 'Username'|trans([], 'admin_ext') }}` or auto fallback with `{{ 'Username'|trans }}`

### DIFF
--- a/public/js/pimcore/element/permissionchecker.js
+++ b/public/js/pimcore/element/permissionchecker.js
@@ -254,7 +254,7 @@ pimcore.element.permissionchecker = Class.create({
         this.resultPanel.removeAll();
 
         var columns = [
-            {text: t("Username"), sortable: false, dataIndex: "userName", editable: false, flex: 150, filter: 'string'}
+            {text: t("username"), sortable: false, dataIndex: "userName", editable: false, flex: 150, filter: 'string'}
         ];
 
         var columnCount = data.columns.length;
--- a/translations/admin.en.yaml
+++ b/translations/admin.en.yaml
@@ -970,6 +970,7 @@ email_log_subject: Subject
 gdpr_data_extractor_label_email: 'Email Address'
 workflow_notes: Notes
 2fa_code: Code
+username: Username
 password: Password
 show_metainfo: Info
 email_log_from: From
--- a/translations/admin_ext.en.yaml
+++ b/translations/admin_ext.en.yaml
@@ -689,7 +689,6 @@ quantityValueRange_field: Quantity Value Range
 numeric: Number
 structuredtable_type_number: Number
 objectsMetadata_type_number: Number
-username: Username
 localized_fields: Localized Fields
 field_collections: Field-Collections
 special_settings_tooltip: Special Settings


kingjia90:
Looks quite inconsistent,

- difference in translation eg. `Passwort` and `Kennwort` when changing to ucfirst
- `{{ 'Username'|trans([], 'admin') }}`  doesn't work, but ` {{ 'Username'|trans([], 'admin_ext') }} ` do? aren't these translation domains merged somewhere?
- Changing to `ucfirst` could be breaking change, if anybody provided translation by their own to overcome these inconsistencies. (or use different naming, eg. "member id")


image
```
    <b>lc admin</b> {{ 'username'|trans([], 'admin') }} <br>
    <b>lc admin</b> {{ 'password'|trans([], 'admin') }} <br>
    <b>cc admin</b> {{ 'Username'|trans([], 'admin') }} <br>
    <b>cc admin</b> {{ 'Password'|trans([], 'admin') }} <br>
    <b>cc admin_ext</b> {{ 'Username'|trans([], 'admin_ext') }} <br>
    <b>cc admin_ext</b> {{ 'Password'|trans([], 'admin_ext') }} <br>
```

dvesh3:
@kingjia90 Yes, It is merged on index https://github.com/pimcore/admin-ui-classic-bundle/blob/e9ad0daae2e9a75f9190eae15e17b98758ee85a3/templates/admin/index/index.html.twig#L617 
https://github.com/pimcore/admin-ui-classic-bundle/blob/433f3241718c4afd39b57416d8b26c336b764300/src/Controller/Admin/MiscController.php#L116
so on login it is still missing. 

I would go with your fix as it is consistent with password. Also, we need to copy the translations from `.ext Username` to POEditor and then import it from there. Maybe we can kick the Uppercase key from `.ext` then as i see it is only used here https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/pimcore/element/permissionchecker.js#L257

WDYT?

translations/admin_ext.fr.yaml
```diff
@@ -174,7 +174,7 @@ all_roles: "Tous les r\xF4les"
 add_role: "Ajouter un r\xF4le"
 role_creation_error: "Le r\xF4le n'a pas \xE9t\xE9 cr\xE9\xE9"
 workspaces: Zone de travail
-Username: Login
+Username: Nom d'utilisateur
```
kingjia90: @mcassier31 Is it right that username can be translated to `login` in French?

kingjia90:
@dvesh3 Ok, so i've copied the translation `Username` from `admin_ext` to `ext` and lowercased the translation key. 
I've also noticed that untranslated lowercase `username` was also present in `admin_ext` , so  safely removed that. 

Also changed the occurency in `permissionchecker` so that internally we don't use it, but kept ucfirst `Username` in ext, maybe is still in use in custom implementations 🤔 what would be the process to "deprecate" it?

translations/admin_ext.fr.yaml
mcassier31: @kingjia90 Not sure if you still need this, but "username" is "nom d'utilisateur" :)

kingjia90: yes, that helped, thank you! 
I've changed it directly on [POEditor](https://poeditor.com/join/project/XliCYYgILb) instead of here

PR #342, opened 2023-11-08
[DataObject] Grid Export - Catch unable to read file exception

dvesh3:


kingjia90 (changes requested):
LGTM, but the same fix for Assets would be missing
https://github.com/pimcore/admin-ui-classic-bundle/blob/2657473c03e0277d2dd56ec2bad3eebe5b09908f/src/Controller/Admin/Asset/AssetHelperController.php#L758

dvesh3:
> LGTM, but the same fix for Assets would be missing
> 
> https://github.com/pimcore/admin-ui-classic-bundle/blob/2657473c03e0277d2dd56ec2bad3eebe5b09908f/src/Controller/Admin/Asset/AssetHelperController.php#L758

@kingjia90 Good catch. Added the try catch for assets too.

PR #347, opened 2023-11-10
[Bug] Avoid multiple save ajax calls

### LINKED ISSUE #346: [Bug]: inconsistent localized values in FieldCollection when hitting save + autosave at the same times
### Pimcore version

10, 11

### Steps to reproduce

If the index of a given field collection gets changed, the autosave will be triggered. If the user saves the object at the same time (which is a jackpot), a race condition will rise up, and the localized field values will be mixed up heavily.

****

In our case, we have a simple field collection with a containing localized field. Because it is very difficult to determinate the exact state, I've made a short video to reproduce it. As you can see, within my second try, I was quick enough to trigger the race condition.

You need to hit ctrl+s instantly after the autosave request has been made.

https://github.com/pimcore/pimcore/assets/700119/a0303742-005c-439a-b753-b4718b157666

### Actual Behavior

Localized values getting mixed up in field collection which leads to corrupt data structure.

### Summary
The values should be stored properly. Since the autosave is just a simple `setTimeout` check, it's a serious issue and occurs randomly to our customers.

### Possible Fix
A simple fix could be to disable the save action (disable save button visually and disable the keyboard shortcut)) in extJs as long the autosave process is ongoing.

kingjia90:
Potentially resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/346

The reasons is that the tab is not being masked during the whole auto-saving process https://github.com/pimcore/admin-ui-classic-bundle/blob/a0b64ffcf41f6bceb8a96297f804fd6c91819ce5/public/js/pimcore/object/object.js#L783-L785 since we cannot mask it on autosave (otherwise it may interrupt editor's work on regular intervals), we briefly notify the user that another save is in progress and to retry later instead, IF, they press the save button or hotkey, exactly during an autosave.

Maybe can refactor be more specific (`isSaving` to `isAutosaving`) that an autosave is in progress, because under normal circumstances without auto-save, the save button and shortcut cannot be spammed.

~`pimcore.helpers.showNotification` parameter `hideDelay` was unused and is also fixed.~

### DIFF
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -88,7 +88,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
         }
 
         if (this.saveInProgress()){
-            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '');
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info");
             return;
         }
 
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -87,6 +87,11 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
             return;
         }
 
+        if (this.isSaving()){
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '', 300);
+            return;
+        }
+
         if(typeof task !== 'string') {
             task = '';
         }
@@ -112,9 +117,12 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                 cancelable: true
             });
 
+            this.saving = true;
+
             const isAllowed = document.dispatchEvent(preSaveDocument);
             if (!isAllowed) {
                 this.tab.unmask();
+                this.saving = false;
                 return false;
             }
 
@@ -187,6 +195,9 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                 }.bind(this),
                 failure: function () {
                     this.tab.unmask();
+                }.bind(this),
+                callback: function (){
+                    this.saving = false;
                 }.bind(this),
             });
         } else {
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -88,7 +88,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
         }
 
         if (this.saveInProgress()){
-            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '', 300);
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '');
             return;
         }
 
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -87,7 +87,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
             return;
         }
 
-        if (this.isSaving()){
+        if (this.saveInProgress()){
             pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '', 300);
             return;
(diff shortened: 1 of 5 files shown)


public/js/pimcore/object/object.js
```diff
@@ -776,6 +775,10 @@ pimcore.object.object = Class.create(pimcore.object.abstract, {
         if (this.tab.disabled || (this.tab.isMasked() && task != 'autoSave')) {
             return;
         }
+        if (this.isSaving()){
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '', 100);
```
robertSt7:
```suggestion
            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '', 300);
```

public/js/pimcore/helpers.js
```diff
@@ -589,6 +589,7 @@ pimcore.helpers.showNotification = function (title, text, type, detailText, hide
             maxWidth: 350,
             closeable: true,
             align: "br",
+            autoCloseDelay: hideDelay || 3000,
```
dvesh3: Do we really need to introduce this prop? as the popup will closed just like other notification.

public/js/pimcore/element/abstract.js
```diff
@@ -244,6 +245,10 @@ pimcore.element.abstract = Class.create({
         }
     },
 
+    isSaving: function(){
```
dvesh3: maybe use more precise name here something like `saveInProgress`?

public/js/pimcore/helpers.js
kingjia90: The parameter hideDelay was already there but unused. 
Looked like a bug/overlook, could split in a separate PR maybe, but noticed it while intending to make these info notification shorter than usual ones, especially since it can be spammed.

dvesh3: thanks for the info. this short delay makes the popup disappear too quick for me and it is then hard to read the message.

kingjia90: what about increasing to 1 second or should i keep standard/default 3 seconds? 
It was shortened on purpose because it's not something that meant to be carefully read, but just a blinking warning, hopefully the initial problem cannot be reproduced so frequently.
I was thinking that somebody may have some weird habits (like myself 😄 ) that spams the save shortcut more than once, then the notification gets stacked

NiklasBr: Windows 11 standard is to show notifications for 5 seconds, I think that is a sensible default.

kingjia90: Resolving, keeping the same default delay.

I've just noticed that the `hideDelay` is meant for the delay before hiding after user mouseover event on the toast notification, and not related to the autocloseDelay, so it would be a wrong fix.

robertSt7 (approved):

public/js/pimcore/document/document.js
```diff
@@ -87,6 +87,11 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
             return;
         }
 
+        if (this.saveInProgress()){
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '');
```
dvesh3:
```suggestion
            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info");
```
4th argument only used for error type

public/js/pimcore/object/object.js
```diff
@@ -776,6 +775,10 @@ pimcore.object.object = Class.create(pimcore.object.abstract, {
         if (this.tab.disabled || (this.tab.isMasked() && task != 'autoSave')) {
             return;
         }
+        if (this.saveInProgress()){
+            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info", '');
```
dvesh3:
```suggestion
            pimcore.helpers.showNotification(t("warning"), t("Another saving process is in progress, please wait and retry again"), "info");
```

dvesh3 (approved):

PR #348, opened 2023-11-13
[Improvement]:  Add download asset name config

robertSt7:
see https://github.com/pimcore/pimcore/pull/16242


robertSt7:
Closed due to https://github.com/pimcore/pimcore/issues/16295

PR #352, opened 2023-11-16
[DataObjects] Do not set general settings data back to object on save

### LINKED ISSUE #351: [Bug]: Class field with name "text" gets overwritten by object key
### Pimcore version

11.1.0

### Steps to reproduce

Create a field called "text" in an object class (input, textarea, wyswyg, number).
Create an object and type something into the "text" field.
Save and publish and reload the object.

### Actual Behavior

The field "text" will be overwritten by the object key.

### Expected Behavior

The field "text" stays on the value that was typed.

This happens at least since version 11.1.0, maybe before that as well.
I tried it with different filed types, at least input, textarea, wyswyig and number are affected. Number throws an error message in the frontend when I type in "1":

Message: Pimcore\Model\DataObject\FAQ::setText(): Argument pimcore/pimcore#1 ($text) must be of type ?float, string given, called in /var/www/html/vendor/pimcore/pimcore/lib/Model/AbstractModel.php on line 186

I assume it immediately writes the key into the field and runs into a validation error.

This does not happen when the "text" field is inside a localizedfields field.

dvesh3:
Resolves #351

## Additional Info
As per debugging, it set the values to object from general settings which are already set.

Also, Object created from API should be same the one created from Admin UI

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectController.php
+++ b/src/Controller/Admin/DataObject/DataObjectController.php
@@ -1381,21 +1381,6 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             }
         }
 
-        // general settings
-        // @TODO: IS THIS STILL NECESSARY?
-        if ($request->get('general')) {
-            $general = $this->decodeJson($request->get('general'));
-
-            // do not allow all values to be set, will cause problems (eg. icon)
-            if (is_array($general) && count($general) > 0) {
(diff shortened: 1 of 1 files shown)


kingjia90:
LGTM, didn't find any negative drawback from the removal.

By comparing the objects, these keys would be different but they don't have a corresponding setter anyways.
image

Merging and let's keep an eye on potential regression,

PR #355, opened 2023-11-17
[Task]: Deprecate date editable format fallback

kingjia90:
Resolves https://github.com/pimcore/pimcore/pull/14431#issuecomment-1441644584

Related https://github.com/pimcore/pimcore/issues/14524
See also https://github.com/pimcore/pimcore/pull/15849

Should be likely removed instead, it's not supported to be passing `strtime` parameters

### DIFF
--- a/public/js/pimcore/document/editables/date.js
+++ b/public/js/pimcore/document/editables/date.js
@@ -30,11 +30,6 @@ pimcore.document.editables.date = Class.create(pimcore.document.editable, {
     render: function () {
         this.setupWrapper();
 
-        if (this.config.format) {
-            // replace any % prefixed parts from strftime format
-            this.config.format = this.config.format.replace(/%([a-zA-Z])/g, '$1');
-        }
-
         if(this.data) {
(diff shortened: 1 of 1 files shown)


public/js/pimcore/document/editables/date.js
```diff
@@ -29,12 +29,14 @@ pimcore.document.editables.date = Class.create(pimcore.document.editable, {
 
     render: function () {
         this.setupWrapper();
+        
+        if (this.config.format && this.config.format.includes('%')) {
```
dvesh3, line 33: @kingjia90 maybe we should also update docs to defined which formats are supported by Extjs? would be nice with a link

dvesh3 (commented):
Please rebase on 1.x as deprecation should go into minor release.

public/js/pimcore/document/editables/date.js
kingjia90: Providing a PR for that https://github.com/pimcore/pimcore/pull/16275

PR #357, opened 2023-11-23
[Bug] Key 32 translated to Space

### LINKED ISSUE #358: [Bug] Key bindings show spacebar as empty string
Adding a keybinding with the spacebar will result in the following UI Problem:

image

For a better user experience we should should change it to "Space" like we do it for Shift, Ctrl and Alt

mattamon:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/358
A Space is not displayed the correct way. 
This replaces the keycode 32 with the Space


dvesh3:
@mattamon I would consider this as a bugfix. fine for you?

mattamon:
@dvesh3 fine for me of course! :)

PR #360, opened 2023-11-24
[Assets] Metadata - Fix xlsx/csv-export does not work

### LINKED ISSUE #359: [BUG]: xlsx/csv-export for assets does not work
### Pimcore version

11.1.0

### Steps to reproduce

I click on Assets -> then on the folder with the images I want to export -> list ->xlsx-export

### Actual Behavior

An ecxel file is downloaded but when I open the file there is only one line with the following values: `[{"key": "preview~system", "label": "preview"},{"key": "id~system", "label": "id"},{"key": "type~system", "label": "type"},{"key": "fullpath~system", "label": "fullpath"},{"key": 'filename~system', "label": "filename"},{"key": "creationDate~system", "label": "creationDate"},{"key": "modificationDate~system", "label": "modificationDate"},{"key": "size~system", "label": "size"}]"`

### Expected Behavior

It should give me a list of all assets with the desired fields

dvesh3:
Resolves #359

### DIFF
--- a/public/js/pimcore/asset/gridexport/xlsx.js
+++ b/public/js/pimcore/asset/gridexport/xlsx.js
@@ -32,8 +32,8 @@ pimcore.asset.gridexport.xlsx = Class.create(pimcore.element.gridexport.abstract
                     fieldLabel: t('header'),
                     name: 'header',
                     store: [
-                        ['name', t('name')],
-                        ['title', t('title')],
+                        ['name', t('system_key')],
+                        ['title', t('label')],
                         ['no_header', t('no_header')]
                     ],
                     labelWidth: 200,
--- a/public/js/pimcore/asset/gridexport/xlsx.js
+++ b/public/js/pimcore/asset/gridexport/xlsx.js
@@ -22,6 +22,26 @@ pimcore.asset.gridexport.xlsx = Class.create(pimcore.element.gridexport.abstract
 
     getDownloadUrl: function(fileHandle) {
          return Routing.generate('pimcore_admin_asset_assethelper_downloadxlsxfile', {fileHandle: fileHandle});
+    },
+
+    getExportSettingsContainer: function () {
+        return new Ext.form.FieldSet({
+            title: t('export_xlsx'),
+            items: [
+                new Ext.form.ComboBox({
+                    fieldLabel: t('header'),
+                    name: 'header',
+                    store: [
+                        ['name', t('name')],
+                        ['title', t('title')],
+                        ['no_header', t('no_header')]
+                    ],
+                    labelWidth: 200,
+                    value: 'title',
+                    forceSelection: true,
+                })
+            ]
+        });
     }
 });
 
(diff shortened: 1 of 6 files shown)


public/js/pimcore/asset/gridexport/xlsx.js
```diff
@@ -22,6 +22,26 @@ pimcore.asset.gridexport.xlsx = Class.create(pimcore.element.gridexport.abstract
...
+    getExportSettingsContainer: function () {
+        return new Ext.form.FieldSet({
+            title: t('export_xlsx'),
+            items: [
+                new Ext.form.ComboBox({
+                    fieldLabel: t('header'),
+                    name: 'header',
+                    store: [
+                        ['name', t('name')],
+                        ['title', t('title')],
```
kingjia90: suggesting to refactor `name` to something like `system name` or `key`, while `title` to `label`

src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -758,6 +759,11 @@ public function doExportAction(Request $request): JsonResponse
             stream_copy_to_stream($fileStream, $temp, null, 0);
 
             $firstLine = true;
+            if ($request->get('initial') && $header === 'no_header') {
+                array_shift($csv);
```
kingjia90:
```suggestion
```

this is uneccessary, because it's already considered by `getCsvData` on
https://github.com/pimcore/admin-ui-classic-bundle/pull/360/files#diff-4547f82eae3e18cc6e0476109b6312b1ed287fc51ad77ba2e9f9e7a7f18c1dd4R821
and the header line won't be pushed to `$csv`

current state it would export n-1 items when selecting no title

public/js/pimcore/asset/gridexport/xlsx.js
kingjia90: but overall, for assets, not sure if it's worth it to provide the 2 options, the only difference is that the first returns the `~system`, the rest is the same, the asset metadata do not have a label system where you can provide a pretty name.

kingjia90 (changes requested):
Just an issue and an input. 
The rest LGTM

public/js/pimcore/asset/gridexport/xlsx.js
dvesh3: I think it makes more sense when you use labels on transformers/formaters and can decide if you want to export name or label. If we still refactor the options then we need to do it for data object exports as well.

kingjia90: I'd say let's change for data object as well, `name` and `title` are quite confusing to me in that context of grid list, at least felt confused without having this picture/screenshot in mind
image
, it was even harder to figure it out for assets.

kingjia90: > I think it makes more sense when you use labels on transformers/formaters and can decide if you want to export name or label. 

but didn't see transformers/formatters for assets, is it just we do not have built-in (examples) one on demo but is/should be supported?

PR #361, opened 2023-11-28
[Improvement] - Dynamic Option Provider

procnet0:
Add context for Blocks/bricks/fieldcollection for select and multiselect option provider.

### DIFF
--- a/public/js/pimcore/object/tags/block.js
+++ b/public/js/pimcore/object/tags/block.js
@@ -311,6 +311,24 @@ pimcore.object.tags.block = Class.create(pimcore.object.tags.abstract, {
 
         this.dataFields = {};
         this.currentData = {};
+
+        this.updateBlockIndices();
+    },
+    updateBlockIndices: function() {
+        for (var itemIndex = 0; itemIndex < this.component.items.items.length; itemIndex++) {
+            var item = this.component.items.items[itemIndex];
+
+            for (j = 0; j < this.currentElements.length; j++) {
+                if (item !== this.currentElements[j].container) continue;
+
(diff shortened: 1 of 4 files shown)


procnet0:

procnet0:
recheck

procnet0:
recheck

robertSt7:
@procnet0 Could you please replace `var` with `let` and `const`. Thank you

mattamon:
> @procnet0 Could you please replace `var` with `let` and `const`. Thank you

@robertSt7 could you have a look again please. I changed it now to `let`

public/js/pimcore/object/tags/block.js
```diff
@@ -311,6 +311,25 @@ pimcore.object.tags.block = Class.create(pimcore.object.tags.abstract, {
...
+    updateBlockIndices: function() {
+        for (let itemIndex = 0; itemIndex < this.component.items.items.length; itemIndex++) {
+            let item = this.component.items.items[itemIndex];
+
+            for (let j = 0; j < this.currentElements.length; j++) {
+                if (item !== this.currentElements[j].container) continue;
+
+                let fields = this.currentElements[j].fields;
+                for (let fieldName in fields) {
+                    if (this.currentElements[j].fields.hasOwnProperty(fieldName)) {
```
robertSt7: @procnet0 Sorry, I don't get why we need this check, because for me `this.currentElemenst[j].fields` is the same as `fields` and you iterate over fields. So the prop `fieldName` should always be in `fields` right?

robertSt7:
@procnet0 Thanks for the improvement

PR #362, opened 2023-11-28
Password reset - fix for possible host header injection

brusch:


src/Controller/Admin/LoginController.php
```diff
@@ -226,12 +234,20 @@ public function lostpasswordAction(Request $request, CsrfProtectionHandler $csrf
...
                 $token = Authentication::generateTokenByUser($user);
 
-                $loginUrl = $this->generateUrl('pimcore_admin_login_check', [
-                    'token' => $token,
-                    'reset' => 'true',
-                ], UrlGeneratorInterface::ABSOLUTE_URL);
-
                 try {
+                    $domain = SystemSettingsConfig::get()['general']['domain'];
+                    if(empty($domain)) {
```
blankse: The variable is always set. It doesn't need empty().
```suggestion
                    if (!$domain) {
```

src/Controller/Admin/LoginController.php
```diff
@@ -226,12 +234,20 @@ public function lostpasswordAction(Request $request, CsrfProtectionHandler $csrf
...
-                $loginUrl = $this->generateUrl('pimcore_admin_login_check', [
-                    'token' => $token,
-                    'reset' => 'true',
-                ], UrlGeneratorInterface::ABSOLUTE_URL);
-
                 try {
+                    $domain = SystemSettingsConfig::get()['general']['domain'];
+                    if (!$domain) {
+                        throw new \Exception('No main domain set in system settings, unable to generate reset password link');
+                    }
```
kingjia90, line 241: LGTM, but there's one potential scenario where one doesn't set the main domain on purpose because has multi-sites set in place. What about to change to an allow list approach? 

Something like
```suggestion
                    $allowedDomains = SystemSettingsConfig::get()['general']['domain'];
                    
                    $siteListing = new Site\Listing;
                    $sites = $siteListing->getSites();
                    foreach ($sites as $site) {
                        $allowedDomains[] = $site->getMainDomain();
                    }
                    
                    $context = $router->getContext();
                    $host = $context->getHost();
                    
                    if (!in_array($host, $allowedDomains)) {
                        throw new \Exception('Unable to generate reset password link: No main domain set in system settings or invalid Host');
                    }
```

maybe even push `localhost` to `$allowedDomains` when debug mode is detected or so?

brusch: @kingjia90 thanks for your feedback. I'd see this as a first fix, even when having multiple sites it would be possible (it's even recommended) to specify a main domain. This could be a domain that's used as a site as well -> that works 😉 

I don't think it makes sense to include all the domains from the site, because usually you don't access the admin UI using a sites domain. 

Let's see if this causes confusion, if so we could extend the current fix. 😊

kingjia90: yes, that sounds fine, merged and tagged.


> I don't think it makes sense to include all the domains from the site, because usually you don't access the admin UI using a sites domain.

Agree, but there's a gray area, it's seem working seamlessly and be supported, at the same time it is not strictly direcommended nor forbidded to access to admin UI from sub-site domain (eg. by force redirecting to main domain admin), even `no-reply` emails are sent with the current site host first https://github.com/pimcore/pimcore/blob/512cc910427e8603572508da1d782b8c6a0ebb5a/lib/Tool.php#L302
https://github.com/pimcore/pimcore/blob/7c50b9c530bcb106e192d7dfd3c34bb5eeaaf038/lib/Mail.php#L499-L502
and one can log-in in two different tabs on same browser window.

If i am not wrong, before PR, reset password request from 
- main site `wvgroup.car/admin` would send an email from `no-reply@wvgroup.car` with a link `wvgroup.car/admin/reset?token=1234` 
and 
- sub-site `sportsche.car/admin` would send an email from `no-reply@sportsche.car` with a link `sportsche.car/admin/reset?token=1234`

after PR, the `sportsche.car/admin` would send an email from `no-reply@sportsche.car` BUT the link would look like `wvgroup.car/admin/reset?token=1234`, ONLY if `system > general > domain` config is set as main site, and if you change it to `sportsche.car`, it's changed for every site.

PR #364, opened 2023-11-30
[Bug]: Relation field wrong dirty detector and "losing" data even without any changes

robertSt7:


public/js/pimcore/asset/metadata/grid.js
```diff
@@ -246,6 +246,9 @@ pimcore.asset.metadata.grid = Class.create({
                         //enable different editors per row
                         editor.editors.each(function (e) {
                             try {
+                                if (e.config?.fieldInfo?.layout?.fieldtype === 'manyToManyRelation') {
```
kingjia90:
```suggestion
                                if (e.fieldInfo?.layout?.fieldtype === 'manyToManyRelation') {
```
seems enough.

kingjia90 (changes requested):
LGTM, even though is quite hardcoded, should be a custom property or so.

- should it be something like `clearBeforeDestroy`
image and check the value here?

- If you open and save without any changes, it would still flag it as dirty field

robertSt7:
see https://github.com/pimcore/asset-metadata-class-definitions/pull/155

public/js/pimcore/element/helpers/gridCellEditor.js
```diff
@@ -149,7 +149,8 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
             startValue = me.startValue,
             value;
 
-        if(fieldInfo.layout.noteditable) {
+        if(fieldInfo.layout.noteditable ||
+        JSON.stringify(this.initValue) === JSON.stringify(this.getValue())) {
```
kingjia90: `initValue` looks undefined.

public/js/pimcore/element/helpers/gridCellEditor.js
```diff
@@ -121,7 +117,11 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
                     text: t("save"),
                     iconCls: 'pimcore_icon_save',
                     handler: function() {
-                        var newValue = tag.getCellEditValue();
+                        if (typeof tag.isDirty === 'function' && tag.isDirty() === false) {
```
kingjia90, line 120: It looks more like `getValue` is empty on `beforeedit`, wouldn't introduce the `isDirty` if it is not already set for all the other tags

kingjia90 (changes requested):

public/js/pimcore/element/helpers/gridCellEditor.js
kingjia90: same as `me.startValue`, could be the bug somewhere there?

robertSt7: This is a left over.

public/js/pimcore/element/helpers/gridCellEditor.js
robertSt7: @kingjia90 It's the same as in the admin-ui https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/pimcore/asset/metadata/tags/manyToOneRelation.js#L198 I thought it would be good to have the same functionality in every tag where it's needed

PR #366, opened 2023-12-05
[Bug]: Add multiselect to avoid `cancel` button clearing selected values

kingjia90:
Followup https://github.com/pimcore/admin-ui-classic-bundle/pull/364, similar problem https://github.com/pimcore/asset-metadata-class-definitions/issues/151 affecting multiselect
noticedd when fixing https://github.com/pimcore/asset-metadata-class-definitions/pull/157

### DIFF
--- a/public/js/pimcore/asset/metadata/grid.js
+++ b/public/js/pimcore/asset/metadata/grid.js
@@ -246,7 +246,7 @@ pimcore.asset.metadata.grid = Class.create({
                         //enable different editors per row
                         editor.editors.each(function (e) {
                             try {
-                                let fieldType = e.fieldInfo?.layout?.fieldtype;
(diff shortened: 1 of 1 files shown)


public/js/pimcore/asset/metadata/grid.js
```diff
@@ -246,7 +246,8 @@ pimcore.asset.metadata.grid = Class.create({
                         //enable different editors per row
                         editor.editors.each(function (e) {
                             try {
-                                if (e.fieldInfo?.layout?.fieldtype === 'manyToManyRelation') {
+                                let fieldType = e.fieldInfo?.layout?.fieldtype;
```
robertSt7:
```suggestion
                                const fieldType = e.fieldInfo?.layout?.fieldtype;
```

PR #371, opened 2023-12-11
Fix: Default data in pimcore_table are not applied

### LINKED ISSUE #353: Default data in pimcore_table are not applied
### Pimcore version:
11.1.1

### Steps to reproduce:
Create a Table Editable:
```twig
{{ pimcore_table("productProperties", {
    "width": 700,
    "height": 400,
    "defaults": {
        "cols": 2,
        "rows": 3,
        "data": [
            ["Attribute name", "Value"],
            ["Color", "Black"],
            ["Size", "Large"],
            ["Availability", "Out of stock"]
            ]
        }
    })
}}
```

### Actual Behavior:
The default data are not taken into account.

### Possible solution:

The initialize method of pimcore.document.editables.table checks if data is not set to apply the default data:

https://github.com/pimcore/admin-ui-classic-bundle/blob/3c7f80d32129837d42a8b198a1e0de5daa210f79/public/js/pimcore/document/editables/table.js#L23

But this check fails, because data is an empty array. Changing ```if (!data) {``` to ```if (!data || data.length === 0) {``` would fix this issue in my case.

fuw-nielsjensen:
Adds a check if data is an empty array. Fixes: #353

### DIFF
--- a/public/js/pimcore/document/editables/table.js
+++ b/public/js/pimcore/document/editables/table.js
@@ -20,7 +20,7 @@ pimcore.document.editables.table = Class.create(pimcore.document.editable, {
     initialize: function($super, id, name, config, data, inherited) {
         $super(id, name, config, data, inherited);
 
-        if (!data) {
+        if (!data || data.length === 0) {
             data = [
                 [" "]
             ];


fuw-nielsjensen:

PR #372, opened 2023-12-11
Fix: gridview filtering with multiple filters starting with null or empty filter

opreacalina90:
Code update: (Empty filter or NULL) can always be set in combination with other filters, regardless of the sequence. This is related to the order of execution of the boolean operators.


opreacalina90:

robertSt7:
@opreacalina90 Thanks for the PR

PR #375, opened 2023-12-20
[Bug]: Removes `Filter by relation` from reverse object relation grid view

### LINKED ISSUE #16273: [Bug, EC]: Grid View filtering: No results using reverse object relations on M2M relations
### Pimcore version

11

### Steps to reproduce

Using Pimcore demo:

- Open grid view for Data Objects/Product Data/Categories
- Add 'Cars' relation to selected grid columns
- Apply
- Select 'Filter by relations' at 'Cars' column head
- Select 'Search objects'
- Select object '/Product Data/Cars/citroen/2CV'
- Apply filter

### Actual Behavior

Existing relation isn't displayed. 

### Expected Behavior

Existing relations should be displayed, too.

kingjia90:
Resolves https://github.com/pimcore/pimcore/issues/16273#issuecomment-1863949815

Removing due to being not supported yet.

The check is on if this function `getRelationFilter` exists
https://github.com/pimcore/admin-ui-classic-bundle/blob/5f9a715c64b8f3e033523b3d07b99930a1689ea2/public/js/pimcore/element/helpers/gridColumnConfig.js#L355

### DIFF
--- a/public/js/pimcore/object/tags/reverseObjectRelation.js
+++ b/public/js/pimcore/object/tags/reverseObjectRelation.js
@@ -20,7 +20,7 @@ pimcore.object.tags.reverseObjectRelation = Class.create(pimcore.object.tags.man
     pathProperty: "fullpath",
 
     getGridColumnConfig: function ($super, fieldConfig) {
-        let gridColumnConfig = $super(fieldConfig);
+        const gridColumnConfig = $super(fieldConfig);
         delete gridColumnConfig.getRelationFilter;
         return gridColumnConfig;
     },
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/reverseObjectRelation.js
```diff
@@ -19,6 +19,12 @@ pimcore.object.tags.reverseObjectRelation = Class.create(pimcore.object.tags.man
 
     pathProperty: "fullpath",
 
+    getGridColumnConfig: function ($super, fieldConfig) {
+        let gridColumnConfig = $super(fieldConfig);
```
robertSt7:
```suggestion
        const gridColumnConfig = $super(fieldConfig);
```

PR #382, opened 2024-01-08
Support displaying localized fields in field collections in version view

BlackbitDevs:
This PR adds support for displaying localized fields within field collections in the version view instead of just displaying "LOCALIZED FIELDS"


robertSt7:
@BlackbitDevs Thanks a lot for the PR

PR #387, opened 2024-01-10
[Improvement] Adding post menu build event

mattamon:
See title

### DIFF
--- a/public/js/pimcore/layout/toolbar.js
+++ b/public/js/pimcore/layout/toolbar.js
@@ -390,7 +390,7 @@ pimcore.layout.toolbar = Class.create({
 
                  var systemItems = [];
 
-                 if (perspectiveCfg.inToolbar("extras.systemtools.requirements")) {
+                 if (perspectiveCfg.inToolbar("extras.systemtools.requirements")) {removeDatapreMenuBuild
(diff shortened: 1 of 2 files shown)


public/js/pimcore/layout/toolbar.js
```diff
@@ -942,6 +942,15 @@ pimcore.layout.toolbar = Class.create({
          // building the html markup for the main navigation
          pimcore.helpers.buildMainNavigationMarkup(menu);
 
+         // Full menu can be checked here
+         const postMenuBuild = new CustomEvent(pimcore.events.postMenuBuild, {
```
kingjia90: IMHO this spot is too "early" to be called `postMenuBuild`, which sounds more for something that should be happening after all the submenus are also built
https://github.com/pimcore/admin-ui-classic-bundle/pull/387/files#diff-b804c1370a5785827a2dbd64fba0c17e9f5f66526531fe54116207175c8b40c2R958

If needed in this spot, i'd rename as `postMainNavigationBuild`, `postRootMenuBuild` or so.

mattamon: Moved it to be the very last step in the function. So it is truly post menu build

PR #388, opened 2024-01-10
Support displaying field collections in the grid view (JS)

BlackbitDevs:
related PHP code
https://github.com/pimcore/pimcore/pull/16444


fashxp:
closed in favor of the alternative approach in #663

PR #389, opened 2024-01-11
[Improvement] Case insensitive search for object title and key in filter grid view

### LINKED ISSUE #6954: Improvement - No case sensitive and wildcard usage in filter grid view
Hi,

At the moment when using a filter in an input column this is case sensitive so you have to search for exact term.
This is not very user friendly. Can this be changed? 

Also using wildcards are not working in this filter like *. Can this also be added?

robin-ntriga:
This solves the following issue:
https://github.com/pimcore/pimcore/issues/6954

This improvement will make sure users can search the grid view Key and Title fields, without having to exactly match the case of the text.

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -336,9 +336,9 @@ class GridHelperService
                     } elseif (in_array($filterField, $systemFields)) {
                         // system field
                         if ($filterField == 'fullpath') {
-                            $conditionPartsFilters[] = 'concat(`path`, `key`) ' . $operator . ' ' . $db->quote('%' . $filter['value'] . '%');
+                            $conditionPartsFilters[] = 'concat(lower(`path`), lower(`key`)) ' . $operator . ' lower(' . $db->quote('%' . $filter['value'] . '%') . ')';
                         } elseif ($filterField == 'key') {
-                            $conditionPartsFilters[] = '`key` ' . $operator . ' ' . $db->quote('%' . $filter['value'] . '%');
+                            $conditionPartsFilters[] = 'lower(`key`) ' . $operator . ' lower(' . $db->quote('%' . $filter['value'] . '%') . ')';
                         } elseif ($filterField == 'id' && $operator !== 'in') {
                             $conditionPartsFilters[] = 'oo_id ' . $operator . ' ' . $db->quote($filter['value']);
                         } elseif ($filterField == 'id' && $operator === 'in') {
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -335,10 +335,12 @@ class GridHelperService
                         }
                     } elseif (in_array($filterField, $systemFields)) {
                         // system field
+                        $lowerCasedFilterValue = strtolower($filter['value']); // lowercase for case insensitive search
+                        $lowerCasedFilterValue = str_replace('*', '%', $lowerCasedFilterValue); // replace wildcard
                         if ($filterField == 'fullpath') {
-                            $conditionPartsFilters[] = 'concat(lower(`path`), lower(`key`)) ' . $operator . ' lower(' . $db->quote('%' . $filter['value'] . '%') . ')';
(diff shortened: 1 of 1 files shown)


robin-ntriga:

robin-ntriga:
recheck

src/Helper/GridHelperService.php
```diff
@@ -336,9 +336,9 @@ public function getFilterCondition(string $filterJson, ClassDefinition $class, ?
                     } elseif (in_array($filterField, $systemFields)) {
                         // system field
                         if ($filterField == 'fullpath') {
-                            $conditionPartsFilters[] = 'concat(`path`, `key`) ' . $operator . ' ' . $db->quote('%' . $filter['value'] . '%');
+                            $conditionPartsFilters[] = 'concat(lower(`path`), lower(`key`)) ' . $operator . ' lower(' . $db->quote('%' . $filter['value'] . '%') . ')';
```
```suggestion
                           // add the following variable to line 337 or outer part of the IF block so that it could be used in multiple places of the code and needs to be changed once
                            $lowerCasedFilterValue = strtolower($filter['value']);
                            
                            
                            // this should be slightly more efficient, as it lowercasing once for the concat and no need to lowecase the $filter['value'] multiple times nor in the DB.
                            $conditionPartsFilters[] = 'lower(concat(`path`, `key`)) ' . $operator .' ' . $db->quote('%' . lowerCasedFilterValue . '%');
```

However, i have some concern that we are only applying these changes to `key` and `fullpath`, it would end up to be inconsistent. 

The best would be instead something general (covering all the fields) and "toggle-able", eg. with a checkbox near the search term box to decide whether it should be case sensitive or not.

Also the issue was also about the wildcard `*`, so we might need to replace `*` with `%` in `$filter['value']` eg. searching for "AU*IA" would result in `LIKE "%AU%IA%"` (when searching for Austria and Australia 😄 )

robin-ntriga: Thank you for the feedback! I pushed a change with:
- Suggested refactor
- lowercasing for the value is done once in PHP `strtolower()` instead of with SQL `lower()`
- Wildcard (*) support

> However, i have some concern that we are only applying these changes to key and fullpath, it would end up to be inconsistent.

I looked into this and it seems a change should be made in the following file in the main Pimcore repository:
https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/ClassDefinition/Data.php :442
```
        if (!$isNumeric) {
            if ($value === 'NULL') {
                if ($operator === '=') {
                    $operator = 'IS';
                } elseif ($operator === '!=') {
                    $operator = 'IS NOT';
                }
            } elseif (!is_array($value) && !is_object($value)) {
                $lowerCasedValue = strtolower($value);
                if ($operator === 'LIKE') {
                    $key = strtolower($key); // case insensitive search
                    $value = str_replace('*', '%', $lowerCasedValue); // wildcard support
                    $value = $db->quote('%' . $value . '%');
                } else {
                    $value = $db->quote($lowerCasedValue);
                }
            }
        }
```
This would add case insensitivity + support for wildcards for the regular text-type fields. If this looks OK to you I can make a PR there also.

> The best would be instead something general (covering all the fields) and "toggle-able", eg. with a checkbox near the search term box to decide whether it should be case sensitive or not.

I am willing to try this, but could not locate the correct files/templates to change the UI for this. Any pointers to the correct files or changes to be made would be welcome.

robertSt7: @robin-ntriga  Thanks for implementing the feedback. 
We would greatly appreciate the PR in the core for all fields.

I think we can skip the toggle-able filter, because this filter is provided by ExtJS https://github.com/pimcore/admin-ui-classic-bundle/blob/e9ad0daae2e9a75f9190eae15e17b98758ee85a3/public/js/pimcore/object/helpers/grid.js#L275 and with your changes it's case insensitive everywhere.

cancan101:
What is the status of the PR?

robertSt7:
@robin-ntriga Thanks for your contribution. I have created an issue for the changes in the core. If you have time we would highly appreciate a pr for that

PR #390, opened 2024-01-11
Prevent accidentical removal of whole folder instead of deleting selected assets / objects

BlackbitDevs:
Theoretically the grid is easy:
- there is a button in the toolbar to delete the whole folder
- and there is the grid where you can select single items, right click and then click *delete* in the context menu to delete the selected items

In practice a lot of users accidentically remove the whole folder although they wanted to delete only some assets / objects. The reason is that they select some items in the grid with the checkboxes and then they click the *Delete* button in the toolbar. After confirming the deletion the whole folder is gone instead of the selected items.

With this PR, is some items are selected and the "Delete folder" button gets clicked, the same behaviour gets used as if the right-click contextmenu > Delete would have been used.
Screenshot 2024-01-11 at 13 09 56

In other words: Only if no items are selected in the grid, the button will delete the whole folder.


BlackbitDevs:

dmytroderkachblackbit:

public/js/pimcore/elementservice.js
```diff
@@ -83,9 +83,19 @@ pimcore.elementservice.deleteElementsComplete = function(options, response) {
 pimcore.elementservice.deleteElementCheckDependencyComplete = function (window, res, options) {
 
     try {
-        let message = res.batchDelete ? t('delete_message_batch') : t('delete_message');
+        let message = '';
+        if (res.batchDelete) {
+            message += sprintf(t('delete_message_batch'), res.itemResults.length) + "<br /><div style='display: block; padding-top: 10px'>";
```
kingjia90:
```suggestion
            message += sprintf(t('delete_message_batch'), res.itemResults.length) + "<br /><div>";
```
`div` tags are already `block`, i'd prefer to let `ul` tag set the paddings (see the other feedback below)

public/js/pimcore/elementservice.js
```diff
@@ -83,9 +83,19 @@ pimcore.elementservice.deleteElementsComplete = function(options, response) {
...
+            res.itemResults.forEach(function (item) {
+                message += "<b style='display: block; text-align: center;'>\"" + htmlspecialchars(item.path + item.key) + "\"</b>";
```
kingjia90: i'd suggest to change to `<li>`, and only make the `item.key` bold, because the path could be the same for each line, also avoiding alignment to center as well, because it's makes more easy to read "vertically", so not too much variation due the line length

```suggestion
                if (res.itemResults.length > 0) {
                    message += "<ul>";
                    res.itemResults.forEach(function (item) {
                        message += '<li>' + htmlspecialchars(item.path) + '<b>' + htmlspecialchars(item.key) + '</b></li>';
                    })
                    message += "</ul>";
                }
```

image

kingjia90 (changes requested):
LGTM so far, found only some html improvement that could be done

kingjia90:
@fashxp wdty? this change makes this `delete` button more flexible. The popup message should be also clear enough to implicitly explain this dual functionality.


Alternatively, we should split in 2 separated buttons (one folder+bin icon eg. combination of 📁❌ and one as delete selection that appears only there's any selection, plus a "clear selection" shortcut )

fashxp:
@kingjia90 I like the idea. We need to make sure though, that the delete button also works if grid not loaded yet - e.g. when grid-tab not opened for assets, or when class selection necessary in objects grid.

public/js/pimcore/elementservice.js
```diff
@@ -83,9 +83,25 @@ pimcore.elementservice.deleteElementsComplete = function(options, response) {
...
+            message += sprintf(t('delete_message_batch'), res.itemResults.length) + "<br /><div>";
+            res.itemResults.forEach(function (item) {
+                if (res.itemResults.length > 0) {
+                    message += "<ul>";
+                    res.itemResults.forEach(function (item) {
+                        message += '<li>' + htmlspecialchars(item.path) + '<b>' + htmlspecialchars(item.key) + '</b></li>';
+                    })
+                    message += "</ul>";
+                }
+            })
```
kingjia90: the suggestion previously applied wasn't right, the `res.itemResults` would be iterated esponentially

```suggestion
            if (res.itemResults.length > 0) {
                message += "<ul>";
                res.itemResults.forEach(function (item) {
                    message += '<li>' + htmlspecialchars(item.path) + '<b>' + htmlspecialchars(item.key) + '</b></li>';
                })
                message += "</ul>";
            }
```

public/js/pimcore/object/folder.js
```diff
@@ -214,10 +214,19 @@ pimcore.object.folder = Class.create(pimcore.object.abstract, {
...
                 scale: "medium",
-                handler: this.remove.bind(this)
+                handler: function () {
+                    var options = this.search.onRawDeleteSelectedRows();
+                    if (!options) {
+                        options = {
+                            "elementType" : "object",
+                            "id": this.id
+                        };
+                    }
```
kingjia90, line 309:
```suggestion
                    var options = this.search.onRawDeleteSelectedRows();
                    if (!options) {
                        options = {
                            "elementType" : "object",
                            "id": this.id
                        };
                    }
```


Also checked the case that Fash mentioned and indeed it doesn't work because the `selectionModel` is not initialized
image

eg. when trying to delete a folder with multiple object types and not having any object type selected, when deleting an image folder when on `preview gallery` mode (not switched to `list` view)
image

This code should be "tweaked", only try to delete the selected lines when found a row selection.

kingjia90: Any news on this?

public/js/pimcore/asset/listfolder.js
```diff
@@ -459,6 +459,25 @@ pimcore.asset.listfolder = Class.create(pimcore.asset.helpers.gridTabAbstract, {
         return config;
     },
 
+    onRawDeleteSelectedRows: function () {
+        var ids = [];
```
kingjia90: another issue i see is that all `var` (at least newly introduced) should be either `let`/`const` eg.
```suggestion
        let ids = [];
```

herbertroth: var, let, and const are used to declare variables, each with different scope, hoisting, and reassignment behaviors:

var: Function or globally scoped, hoisted, can be redeclared and reassigned. Generally avoided in modern JS due to less predictable scope.
let: Block-scoped, hoisted but not accessible until declared, can be reassigned but not redeclared in the same scope. Preferred for variables that change.
const: Block-scoped, hoisted but not accessible until declared, cannot be reassigned or redeclared. Use for variables that should not change after initialization.
Usage Guidelines:

Avoid var in favor of let and const to prevent scope-related bugs.
Use let for variables that need to be reassigned.
Default to const for variables that should not change, enhancing code readability and safety.

kingjia90:
Do you need any further info on this? Let me know

public/js/pimcore/object/folder.js
BlackbitDevs: Have added a check if the grid is actually loaded. If not, then the current behaviour for the button gets used (delete folder).

mattamon:
@kingjia90 any update on this? can we continue here?

kingjia90:
Thank you both for the changes and the reminder! 
LGTM now, it would be just missing a `CHANGELOG.md` entry to announce this improvement, then we can merge it

kingjia90 (approved):

BlackbitDevs:
@kingjia90 Changelog entry added.

PR #391, opened 2024-01-11
[bugfix] Do not call htmlspecialchars with null

Daniel-Ateles:
En legacy environments stuff can sometimes be null which htmlspecialchars doesn't like. Hence this PR to avoid that.


fashxp:
can you please rebase to 1.3 as this is a bugfix. thx very much!

Daniel-Ateles:
> can you please rebase to 1.3 as this is a bugfix. thx very much!

Aha, I'm sorry. Didn't know that. If I understand correctly it looks like kngjia90 have made this change after your request?

kingjia90:
@Daniel-Ateles yes, luckily it didn't need any changes on the branch itself and a base branch change on Git was enough

fashxp:
thx very much!!

PR #393, opened 2024-01-15
Refactor export handler from inline into own function

Corepex:


mattamon (approved):
LGTM
@kingjia90 wdyt? We would need it in 1.4

kingjia90:
Not sure how it works on ExtJS, but when setting an `id`, it would be "global" and unique, right? 
Does it work properly when you have 2 or more tab with grids open?

My point is that we may need something like a suffix eg. `_grid_2` that identifies the current tab/grid.

Corepex:
> Not sure how it works on ExtJS, but when setting an `id`, it would be "global" and unique, right? Does it work properly when you have 2 or more tab with grids open?
> 
> My point is that we may need something like a suffix eg. `_grid_2` that identifies the current tab/grid.

very good point - didn't think of that - i will check it ...

kingjia90:
By quickly look at the other PR, if we have to apply changes to admin ui classic bundle anyways, shouldn't be easier by refactoring to have this 
https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/pimcore/object/helpers/gridTabAbstract.js#L316-L318
in an own method and call it (eg. `currentGridTab.startExport('csv')`) instead of simulating([?](https://github.com/pimcore/copilot-showcase-bundle/pull/3/files#diff-22ec4f44aac3d603424fd0575b62dbf354779705511d8ee706cbc89f9994a5beR40)) a trigger/click of a button? 

It then also overcome the "limit" of the copilot to be only exporting in CSV format. It could be easily extended to XLSX by passing a different `exportType` parameter. 
https://github.com/pimcore/admin-ui-classic-bundle/blob/a5de401344f711ad49a3fe45da8a431e5e6fe3cf/public/js/pimcore/object/helpers/gridTabAbstract.js#L264
image

Alternatively just could use `getExportButtons()` directly

Corepex:
@kingjia90, thanks for pointing me in the right direction. I moved the export functionality to a separate function (as you mentioned). With that it is now possible to start the export via the `currentlyActiveTab` .... 
This solution feels much more reliable than the id (or class) guessing we had in mind.

kingjia90 (approved):
LGTM, but need to change the PR's title before merging

PR #395, opened 2024-01-15
[BUG, EC] Rendering of localized classification store is slow in backend for many languages

### LINKED ISSUE #394: [BUG, EC] Rendering of localized classification store is slow in backend for many languages
### Discussed in https://github.com/pimcore/pimcore/discussions/16392

<div type='discussions-op-text'>

<sup>Originally posted by **hbauer** December 22, 2023</sup>
In a current project, we need a localizable Classification Store that is displayed in a separate tab in a product-class. In the tab, for example, are 19 groups with 140 attributes from the Classification Store displayed - for 14 languages.
The problem is the poor performance when rendering the attributes because as soon as the tab is clicked, it takes approx. 6-8 seconds until the tab is activated and the content appears. For other products with more attributes it can take 20-30 seconds. Switching between languages takes only 0.5 seconds.

It seems that not only the content of the current language but for all languages is rendered. The more languages are added, the longer it takes. I tried to set "deferredRenderer: true" in the panel config but without any effect.

Tested with the current Pimcore v11.1.4, but the problem also occurs in Pimcore 10.6.

Does anyone else have the same problem or a solution?</div>

robertSt7:
resolves #394

### DIFF
--- a/public/js/pimcore/object/tags/classificationstore.js
+++ b/public/js/pimcore/object/tags/classificationstore.js
@@ -152,21 +152,6 @@ pimcore.object.tags.classificationstore = Class.create(pimcore.object.tags.abstr
                 this.languageElements[this.currentLanguage] = [];
                 this.groupElements[this.currentLanguage] = {};
 
-                var childItems = [];
-
-
-                for (var groupId in this.fieldConfig.activeGroupDefinitions) {
-                    var groupedChildItems = [];
-
-                    if (this.fieldConfig.activeGroupDefinitions.hasOwnProperty(groupId)) {
-                        var group = this.fieldConfig.activeGroupDefinitions[groupId];
-
-                        var fieldset = this.createGroupFieldset(this.currentLanguage, group, groupedChildItems);
-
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/classificationstore.js
```diff
@@ -184,9 +169,28 @@ pimcore.object.tags.classificationstore = Class.create(pimcore.object.tags.abstr
...
+                                if (this.fieldConfig.activeGroupDefinitions.hasOwnProperty(groupId)) {
+                                    const group = this.fieldConfig.activeGroupDefinitions[groupId];
+
+                                    const fieldset = this.createGroupFieldset(this.currentLanguage, group, groupedChildItems);
 
+                                    childItems.push(fieldset);
+                                }
+                            }
+
+                            tab.add(childItems);
```
kingjia90, line 193: This in combination with `activate` seems wrong: if i try to switch tab between the languages, eg. `EN`, `FR`, `EN`, `FR`, it would end up appending the items for each time i click on it.

PR #404, opened 2024-01-23
Add translation for invalid option in multiselect / select

melissakittl:
In a Select/Multiselect were different values available (A, B, C). Option C was chosen for an object but later removed from the option list. Because the select still holds the value C, it triggers an "invalid option" error as C is no longer valid. The Select appears empty, hiding the retained value, and the shown error is meaningless.
image

Added translations for better understanding:
image


robertSt7:
@melissakittl Thank you for the PR

PR #419, opened 2024-02-12
[Bug]: Fix asset metadata tag type

kingjia90:
Related https://github.com/pimcore/asset-metadata-class-definitions/pull/169

### DIFF
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -545,7 +545,7 @@ pimcore.element.helpers.gridColumnConfig = {
             editor.setObject(this.object);
         } else {
             let layoutInfo = this.fieldObject[fieldInfo.dataIndex].layout;
-            const tagType = layoutInfo.fieldConfig?.type ?? this.fieldObject[fieldInfo.dataIndex].type;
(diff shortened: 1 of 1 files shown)


public/js/pimcore/element/helpers/gridColumnConfig.js
```diff
@@ -544,8 +544,8 @@ pimcore.element.helpers.gridColumnConfig = {
             var editor = new pimcore.object.tags[tagType](null, fieldInfo.layout.layout);
             editor.setObject(this.object);
         } else {
-            var tagType = this.fieldObject[fieldInfo.dataIndex].layout.fieldtype;
-            let layoutInfo = this.fieldObject[fieldInfo.dataIndex].layout
+            let layoutInfo = this.fieldObject[fieldInfo.dataIndex].layout;
+            const tagType = this.fieldObject[fieldInfo.dataIndex].type ?? layout.fieldtype;
```
kingjia90, line 548: The console log is
image
Don't know when and where the `fieldtype` was used, maybe a typo but kept as fallback

```suggestion
            const tagType = this.fieldObject[fieldInfo.dataIndex].type;
```
in case, could just removed

robertSt7 (approved):

PR #420, opened 2024-02-12
[IMPROVEMENT] Assets:upload zip archive-option to overwrite files

opreacalina90:
Here is an improvement for avoiding duplicates while uploading a zip with multiple files


cancan101:
What is the status of this pr? 

I noticed that when uploading a folder (through drag and drop), pimcore just duplicates rather than asking if I want to overwrite.

robertSt7:
Hi @opreacalina90 

I checked your PR and I see that you change some non-English translations files.
Actually only the update of the English translations have an effect, because these keys are then synced with POEditor. https://poeditor.com/join/project/XliCYYgILb
You can then translate it there to other languages.

Could you please remove all changes from the non-English files and fix the conflicts?

Thanks

robertSt7:
@opreacalina90 Thanks a lot for the improvement

PR #423, opened 2024-02-15
Added Gotenberg support for creating page and version previews

brusch:
See also https://github.com/pimcore/pimcore/pull/16608

### DIFF
--- a/src/Controller/Admin/Document/DocumentController.php
+++ b/src/Controller/Admin/Document/DocumentController.php
@@ -28,7 +28,7 @@ use Pimcore\Config;
 use Pimcore\Controller\KernelControllerEventInterface;
 use Pimcore\Db;
 use Pimcore\Event\Traits\RecursionBlockingEventDispatchHelperTrait;
-use Pimcore\Image\Chromium;
+use Pimcore\Image\HtmlToImage;
 use Pimcore\Logger;
 use Pimcore\Model\Document;
 use Pimcore\Model\Document\DocType;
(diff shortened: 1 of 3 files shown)


brusch:
Tests will work as soon as pimcore/pimcore:11.2 was released

kingjia90 (approved):

src/Controller/Admin/Document/DocumentController.php
```diff
@@ -1056,8 +1056,8 @@ public function diffVersionsAction(Request $request, int $from, int $to, Documen
         }
 
         try {
-            Chromium::convert($prefix . $router->generate('pimcore_admin_document_document_diff_versions_html', ['id' => basename($fromHtmlFile)]), $fromImageFile);
-            Chromium::convert($prefix . $router->generate('pimcore_admin_document_document_diff_versions_html', ['id' => basename($toHtmlFile)]), $toImageFile);
+            HtmlToImage::convert($prefix . $router->generate('pimcore_admin_document_document_diff_versions_html', ['id' => basename($fromHtmlFile)]), $fromImageFile);
```
kingjia90, line 1059: Ideally, we could create a new method in pimcore `11.2` eg. `convertByHtml` and pass the string `$docContentFrom` and change in the core to use [$page->setHtml()](https://github.com/chrome-php/chrome?tab=readme-ov-file#set-the-page-html) instead of [navigate](https://github.com/pimcore/pimcore/pull/16608/files#diff-74767dcb230d5e59d6f91308ab2f91896c04ea2a15292c88915745eefb12f247R189)

Performance-wise should be faster and avoid temp files

kingjia90: Leaving for future, so far works

PR #431, opened 2024-02-19
[BUG]: Send invitation link always hidden

TobiSW:
### Steps to reproduce

Login with any user in Pimcore 11.x. Some logic changed after this bugfix => https://github.com/pimcore/pimcore/pull/16131. The user last login field is now set properly, which impact the visibility of the sent invitation link to other users.

### Expected Behavior
Users with the correct access rights should be able to always sent invitation links to other users

### Solution idea
Remove the user lastLogin timestamp check.

### DIFF
--- a/public/js/pimcore/settings/user/user/settings.js
+++ b/public/js/pimcore/settings/user/user/settings.js
@@ -241,7 +241,7 @@ pimcore.settings.user.user.settings = Class.create({
                     xtype: "button",
                     style: "margin-left: 8px",
                     iconCls: "pimcore_nav_icon_email",
-                    hidden: (this.currentUser.lastLogin > 0) || (user.id == this.currentUser.id),
+                    hidden: (user.id === this.currentUser.id),
                     handler: function () {
                         Ext.Ajax.request({
                             url: Routing.generate('pimcore_admin_user_invitationlink'),


TobiSW:

TobiSW:
recheck

kingjia90:
I don't get this change, that condition was to make sure that the invitation would be sent only for the first time and never show it anymore once the user have been logged-in at least once. Don't see the utility of sending the invitation multiple times. 

I guess you were using it to kind of reset password link on user behalf, right?

TobiSW:
Yes exactly, in our case some admin users were able to send invite links again. As it was already possible for a long time before that last bugfix with the last login update was introduced, the users were kind of used to this behaviour.

kingjia90:
In that case, i'd say that this PR is not valid as it is not a bug, and that an admin-only button to trigger a reset password (without the need of typing or copying the email) could be a "feature" with a toggle-able permission based on role/user. 

In this way, we can keep the email message be distinct and avoid any potential breaks when changing the invitation behavior. wdyt?

TobiSW:
Iam not sure what is the best way to go here. As there is already a specific permission to handle users/roles. Iam not sure if it needs an extra permission to only show that button inside a user edit. For me the password reset and the invite link are two seperate things. Actually they are seperated in their logic and behaviour. 

I get your point to only show this one time but I dont understand, what could be a potential break here? I think if the invite logic change, you need to check in any case if the initial invite is still working.

If you want to keep the logic to only see the invite button one time when a user was never logged in, understood and totally fine, then we will just communicate with our clients, that this behaviour changed and that they should use the password forgot functionalitiy instead.

kingjia90:
Ops, wrongly thought of the permissions part, was more thinking about a way to enable/disable (depending on own need), but nevermind, it's already in the admin domain. 

Yes, the functionality was never meant to be used that way (to reset password). 
Sorry, didn't meant "break" but more referring to cases like that the message get rephrased and embellished to be more targeted to first time onboarding or so, making it sound weird in the password resetting context.

TobiSW:
Dont worry, totally understood your point, now even more what is the intention behind. Thanks for your feedback on that, helps us also to communicate this better with the client.

PR #437, opened 2024-02-23
Fix csrf token generation in LoginController::loginAction

melissakittl:
# Problem
While working in admin interface in one browser tab, then opening another one via `/admin/login` or a deeplink `admin/login/deeplink?object_29_object`, the csrf token gets regenerated with `$force = true`. If you switch back to the first browser tab, the csrf token is not valid anymore, you get an `Access denied` error and have to reload the admin interface completely.

## Expected behavior
Working on multiple tabs is possible without reloading the admin interface (some changes might get lost otherwise)

## Actual behavior
In the first browser tab you see the following error, when saving an element (or doing something else)
image
In the logs you see: `[2024-02-23T11:03:43.679029+01:00] security.ERROR: Detected CSRF attack on /admin/object/save {"request":"/admin/object/save"} []`

# Changes in this pull request
First check, if there is already a csrf token set in the session. If no csrf token can be found, then regenerate with `$force = true`in `LoginController::loginAction`

### DIFF
--- a/src/Controller/Admin/LoginController.php
+++ b/src/Controller/Admin/LoginController.php
@@ -114,7 +114,10 @@ class LoginController extends AdminAbstractController implements KernelControlle
             return new RedirectResponse($redirectUrl);
         }
 
-        $csrfProtection->regenerateCsrfToken($request->getSession());
+        // check csrf token before generating a new one with force=true
+        if (!$csrfProtection->getCsrfToken($request->getSession())) {
+            $csrfProtection->regenerateCsrfToken($request->getSession());
+        }
 
         $user = $this->getUser();
         if ($user instanceof UserInterface) {


src/Controller/Admin/LoginController.php
```diff
@@ -114,7 +114,10 @@ public function loginAction(
             return new RedirectResponse($redirectUrl);
         }
 
-        $csrfProtection->regenerateCsrfToken($request->getSession());
+        // check csrf token before generating a new one with force=true
+        if (!$csrfProtection->getCsrfToken($request->getSession())) {
```
kingjia90, line 118: Thank you for the PR!
LGTM just maybe
```suggestion
        if (!$csrfProtection->getCsrfToken($request->getSession()) && is_null($request->get('deeplink'))) {
```
as i am not sure it should change the behaviour other than for  the deeplink

melissakittl: @kingjia90 but some users might have bookmarked the `/admin/login` link. So, if they are simultaneously working in the backend in one tab and open another tab via the bookmark, the first tab will throw an Access Denied Exception, which is not the desired outcome.

kingjia90: mhh you're right, that could be a case

markus-moser:
Great, thanks a lot 👍

PR #439, opened 2024-02-26
Don't use 100% width for grid/cell editor

blankse:
We can set fields to width 100% with label top. This is really nice for the edit view:
image

But in the grid view this width shouldn't be set. So the grid columns are too wide:
image


robertSt7:
@blankse Thanks for the fix and for the refactoring :)

PR #441, opened 2024-02-27
Fix label in string contains operator

saschabaecher-twocream:
## Situation:
- Open the Grid View in an object folder
- Open the Grid View configuration
- Add the "Operator String Contains" operator
- Change the label of the operator
- The operator label is not saved! Is always "Operator String Contains"

## Expected Behavior:
Adoption of the changed label after saving in the Grid View.

## Fix:
Add value from the text field "Label" in the node configAttributes as label.


saschabaecher-twocream:

robertSt7:
@saschabaecher-twocream Thanks for the PR

PR #444, opened 2024-02-28
Fix group name in class definition tree

saschabaecher-twocream:
## Situation:
- Open the class definition window
- Create a class definition
- Define a group name that is not stored in the Admin Translation
- Class definition is subgrouped in the "Root" folder

## Expected Behavior:
If the group name has no translation, the class definition should be subgrouped under the specified group name.

## Fix:
It must now be checked here whether there is a translation. If not, the group name should be used.

### DIFF
--- a/src/Controller/Admin/DataObject/ClassController.php
+++ b/src/Controller/Admin/DataObject/ClassController.php
@@ -175,7 +175,11 @@ class ClassController extends AdminAbstractController implements KernelControlle
                 }
             }
 
-            $groupName = Translation::getByKeyLocalized($groupName, Translation::DOMAIN_ADMIN, true, true);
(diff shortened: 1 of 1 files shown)


robertSt7:
Hi @saschabaecher-twocream I am sorry, but I am not able to reproduce this bug. Could you please share some screenshots and describe them? Thanks

saschabaecher-twocream:
> Hi @saschabaecher-twocream I am sorry, but I am not able to reproduce this bug. Could you please share some screenshots and describe them? Thanks

Hi @robertSt7,

I have now tested this in the latest version of Pimcore, and the problem no longer exists! In this version, if the group name is not yet present in the admin translations, it is automatically added, and the corresponding key is then stored for each language.

Therefore the pull request can be closed.

Thank you for your review!

PR #446, opened 2024-02-29
PHP 8.3 Support

blankse:
See https://github.com/pimcore/pimcore/pull/16589


robertSt7:
@blankse Thanks for the contribution

PR #448, opened 2024-03-01
Introduce event to handle different object edit lock tasks

APochmann:
### About this contribution
According to the contribution rules in pimcore README.md file I made an attempt to ask on [Gitter channel](https://app.gitter.im/#/room/#pimcore_pimcore:gitter.im/$9I4ZmPLGg1hlRSCyLg1UlU3l0_tn8wIZ556qpOOXO-k) last June if this might be an interesting feature - without success. CONTRIBUTING.md on the other hand does not explicitly state how to contribute new feature besides doing a fork for pull requests which I try now 😉

## Current Situation
The current implementation has a fix behavior in case of parallel access of objects by users in admin UI. The first user opens the object and it gets a lock entry. The second user is prompted to take over this lock.

My issue is that I need a more flexible implementation to have the possibility to keep the lock and e.g. offer the object for the second user in read only mode.

This PR introduces a new event which is triggered in case of user tries to open a locked object. Based on the event the implementation can decide to

- Offer the dialog as up to now to override the lock
- Keep the existing lock
- Override the lock without asking the user

Additional actions (like ensure read only access) can anyhow be taken at the pre edit event `pimcore.admin.dataobject.get.preSendData`

## Additional info  

New admin event `pimcore.admin.dataobject.get.isLocked` is introduced which supports a parameter `task `with following possible values:

- `TASK_RESPONSE = 'response';`
Default: On active edit lock answer with editlock response as up to now
- `TASK_OVERWRITE = 'overwrite';`
On active edit lock overwrite it with new user
- `TASK_KEEP = 'keep';`
On active edit lock keep existing entry, further actions are up to the implementation

This task will be executed when event returns

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectController.php
+++ b/src/Controller/Admin/DataObject/DataObjectController.php
@@ -64,6 +64,13 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     use DataObjectActionsTrait;
     use UserNameTrait;
 
+    /** On active edit lock answer with editlock response */
+    const TASK_RESPONSE = 'response';
+    /** On active edit lock overwrite with new user */
+    const TASK_OVERWRITE = 'overwrite';
+    /** On active edit lock keep existing entry */
+    const TASK_KEEP = 'keep';
+    
     protected DataObject\Service $_objectService;
 
     private array $objectData = [];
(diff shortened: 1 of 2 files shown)


robertSt7:
Hi @APochmann we very much appreciate your PR, thank you for your contribution

PR #449, opened 2024-03-02
Thumbnail: Don't create cover item with positioning: null

blankse:
If you create a cover item for a thumbnail configuration, the positioning is null. When you save it without changing it. You get following error, when you want create this thumbnail:
```
TypeError: Pimcore\Image\Adapter::cover(): Argument #3 ($orientation) must be of type array|string, null given in /vendor/pimcore/pimcore/lib/Image/Adapter.php:132
```

See also: https://github.com/pimcore/pimcore/pull/16720


fashxp:
thx very much

PR #450, opened 2024-03-02
Fix invite user modal translation

blankse:
Before:
Screenshot 2024-03-02 162659
After:
image


blankse:
Now at least the English text comes. However, I don't know why the German text isn't there.

fashxp:
thx very much.

PR #454, opened 2024-03-05
[Data Objects] Password : Deprecate support for algorithms other than password_hash

### LINKED ISSUE #16294: [Data Objects] Password : Deprecate support for algorithms other than password_hash
### Improvement description

Related to https://github.com/pimcore/pimcore/issues/16292

kingjia90:
Related https://github.com/pimcore/pimcore/pull/16736, https://github.com/pimcore/pimcore/issues/16294


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+#### v1.3.0
```
blankse:
```suggestion
#### v1.4.0
```

public/js/pimcore/object/classes/data/password.js
```diff
@@ -110,7 +110,7 @@ pimcore.object.classes.data.password = Class.create(pimcore.object.classes.data.
             const saltCombo = new Ext.form.field.ComboBox({
                 xtype: "combo",
                 width: 300,
-                fieldLabel: t("saltlocation"),
+                fieldLabel: t("saltlocation") + t("(deprecated)"),
```
blankse: Why not using the existing deprecated translation key?
```suggestion
                fieldLabel: t('saltlocation') + '(' + t('deprecated') + ')',
```

kingjia90: Wanted to use something new that could be hidden entirely, by changing the translation key value to " ", if i use the suggested one with concatenated text, it would be "( )".

Maybe it shouldn't be a matter because it is a field only visible to superusers or more advanced technical users, but in case it gets annoying for UI purposes.

fashxp (approved):

PR #455, opened 2024-03-05
Hide buttons for not editable tables

APochmann:
If a table is readonly in Edit tab of an object in Admin UI it does not make sense to show inactive buttons for table modifications. This small change hides the buttons in this case for having UI without non usable buttons (see example below)

### screenshot without change
image

### screenshot after modification
image

### DIFF
--- a/public/js/pimcore/object/tags/table.js
+++ b/public/js/pimcore/object/tags/table.js
@@ -161,49 +161,54 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
 
         this.cellEditing = Ext.create('Ext.grid.plugin.CellEditing', {});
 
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/table.js
```diff
@@ -161,50 +161,55 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
 
         this.cellEditing = Ext.create('Ext.grid.plugin.CellEditing', {});
 
-        var tbar = [];
+        var tbar = null;
```
robertSt7:
```suggestion
        let tbar = null;
```

robertSt7:
@APochmann Thanks a lot for the improvement

PR #459, opened 2024-03-11
Followup #454

blankse:
https://github.com/pimcore/admin-ui-classic-bundle/pull/454


blankse:
@kingjia90 The translation key `(deprecated)` doesn't exist. I created a follow up PR.

kingjia90:
opss... thank you very much

PR #460, opened 2024-03-12
Document version conparison fix: make the comparison independent from session/cookie authentication

brusch:


src/Controller/Admin/Document/DocumentController.php
```diff
@@ -1088,45 +1099,57 @@ public function diffVersionsAction(Request $request, int $from, int $to): Respon
...
             $result = $image1->compareImages($image2, Imagick::METRIC_MEANSQUAREERROR);
             $result[0]->setImageFormat('png');
 
-            $result[0]->writeImage($diffFile);
+            $result[0]->writeImage($diffImageFile);
             $result[0]->clear();
             $result[0]->destroy();
 
-            $viewParams['image'] = $diffFileId;
+            $viewParams['image'] = base64_encode(file_get_contents($diffImageFile));
```
kingjia90: Overall LGTM, just noticed this part, instead of `writeImage`, could be instead passing an `getImageBlob` to avoid a temp file?

```suggestion
            $viewParams['image'] = base64_encode($result[0]->getImageBlob());
```

kingjia90 (approved):

src/Controller/Admin/Document/DocumentController.php
brusch: @kingjia90 Yep, makes sense - pushed the change -thx

PR #461, opened 2024-03-12
Followup #460

blankse:
https://github.com/pimcore/admin-ui-classic-bundle/pull/460


blankse:
@brusch @kingjia90 I created a follow up PR to #460

kingjia90:
ahh nice 👍🏻

PR #462, opened 2024-03-13
[Bug] JS Errors on Bricks with many-to-many relation and specific configuration

### LINKED ISSUE #406: [Bug] JS Errors on Bricks with many-to-many relation and specific configuration
Hi!

A user creates a new brick in the Admin UI. It contains a FieldContainer and a Fieldset. The fieldset contains an (advanced) many-to-many relation. 
[objectbrick_Test_export.json](https://github.com/pimcore/admin-ui-classic-bundle/files/14062696/objectbrick_Test_export.json)
 You can then also add corresponding objects in the relation. 

Bildschirmfoto 2024-01-26 um 09 34 33

User 2 then has the idea of dragging and dropping the fieldset into the FieldContainer. At that moment, the bricks are no longer displayed, but the browser console is full of JS errors.
Bildschirmfoto 2024-01-26 um 09 34 57
Bildschirmfoto 2024-01-26 um 09 35 19
[objectbrick_Test_export-3.json](https://github.com/pimcore/admin-ui-classic-bundle/files/14062706/objectbrick_Test_export-3.json)

Pimcore 11.1.4 
Admin UI 1.3.1

Best regards
Andreas

robertSt7:
resolves #406

### DIFF
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -3290,14 +3290,31 @@ pimcore.helpers.treeDragDropValidate = function (node, oldParent, newParent) {
         if (disabledLayoutTypes.includes(newParent.data.editor.type)) {
             return false;
         }
+
+        return this.isComponentAsChildAllowed(newParent, node);
     }
 
-    if (newParent.data.root) {
+    if (newParent.data.root && node.data.type !== 'layout') {
         return false;
     }
 
     return true;
 };
+
+pimcore.helpers.isComponentAsChildAllowed = function (parentNode, childNode) {
+    const parentType = parentNode.data.editor.type;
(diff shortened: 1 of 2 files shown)


lukmzig:
@robertSt7 this will not resolve the existing error right? So if there is existing configuration it will still throw an error. (but I guess we cannot really do anything here as this should not be allowed in the first place)
I also tried a drag a `fieldset` from the existing `field container and then back into the `field container` and I got a JS error which then broke the UI. 
Screenshot from 2024-06-14 09-49-45

Can you have a look?

lukmzig (approved):
@robertSt7 seems like issue was only present on my installation. I will merge the PR

PR #465, opened 2024-03-15
464 fix filtering objects or variant in grid tab

### LINKED ISSUE #464: There is no way to filter objects in grid views based on whether they are of variant type or not
I have already posted this issue in "**pimcore/pimcore**" repo. I report in this repo just to have a link to the [issue: #16681](https://github.com/pimcore/pimcore/issues/16681).

### The issue:
**Steps to reproduce**
Select a folder. Now in the grid view you can see objects and variants together.

**Actual Behavior**
In the grid views, all objects in a given folder are retrieved, including variants.
Having removed in Pimcore 11 the ability to do a direct query, I am no longer able to get only the objects or only the variant objects.

**Expected Behavior**
It would help me greatly to have a way to choose which objects to show in the grid view, based on whether they are normal objects, variants, or perhaps both.

### LINKED ISSUE #16681: [Improvement]: There is no way to filter objects in grid views based on whether they are of variant type or not
### Pimcore version

11.1.4

### Steps to reproduce

Select a folder. Now in the grid view you can see objects and variants together. 

### Actual Behavior

I have this problem. 
In the grid views, all objects in a given folder are retrieved, including variants. 
Having removed in Pimcore 11 the ability to do a direct query, I am no longer able to get only the objects or only the variant objects. 
I wonder if there might be a solution. 
Thanks

### Expected Behavior

It would help me greatly to have a way to choose which objects to show in the grid view, based on whether they are normal objects, variants, or perhaps both.

lucfumagalli:
### ISSUE 464: There is no way to filter objects in grid views based on whether they are of variant type or not
This PR is related to issue #464.

I resolved the issue inserting a select ComboBox in the toolbar of objects grid view. Now you can select what type of objects you can show.

### DIFF
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -259,7 +259,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                         this.grid.filters.clearFilters();
                         this.grid.getStore().clearFilter();
 
-                        this.store.getProxy().setExtraParam(selected, true);
+                        this.store.getProxy().setExtraParam("listing", selected);
 
                         this.pagingtoolbar.moveFirst();
 
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -247,6 +247,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                 xtype: "combo",
                 displayField:'name',
                 valueField: "value",
+                hidden: !this.element.data.general.allowInhreitance,
                 store: selectObjectOptions,
                 editable: false,
                 width : 300,
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -259,7 +259,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                         this.grid.filters.clearFilters();
                         this.grid.getStore().clearFilter();
 
-                        this.store.getProxy().setExtraParam("listing", selected);
+                        this.store.getProxy().setExtraParam("filter_by_object_type", selected);
 
                         this.pagingtoolbar.moveFirst();
 
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -234,7 +234,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
             var selectObjectOptions = Ext.create('Ext.data.Store', {
                 fields: ['name', 'value'],
                 data: [
-                    [t("empty"), "all_objects"],
+                    [t("all_types"), "all_objects"],
                     [t("only_object"), "only_objects"],
                     [t("only_variant"), "only_variant_objects"],
                 ]
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -231,6 +231,44 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
 
             this.store.getProxy().setExtraParam("query", this.searchFilter);
 
+            var selectObjectOptions = Ext.create('Ext.data.Store', {
+                fields: ['name', 'value'],
+                data: [
+                    [t("empty"), "all_objects"],
+                    [t("only_object"), "only_objects"],
+                    [t("only_variant"), "only_variant_objects"],
+                ]
+            });
+
+            this.selectObjectType = new Ext.form.ComboBox({
+                fieldLabel: t('select_objects_type'),
+                name: 'objects_type',
+                labelWidth: 120,
+                xtype: "combo",
+                displayField:'name',
+                valueField: "value",
+                store: selectObjectOptions,
+                editable: false,
+                width : 300,
+                triggerAction: 'all',
+                value: 'all_objects',
+                listeners: {
+                    change: function(comboBox,selected){
+                        this.grid.getStore().setRemoteFilter(false);
+                        this.grid.filters.clearFilters();
+                        this.grid.getStore().clearFilter();
+
+                        this.store.getProxy().setExtraParam(selected, true);
+
+                        this.pagingtoolbar.moveFirst();
+
+                        this.grid.getStore().setRemoteFilter(true);
+
+                        this.saveColumnConfigButton.show();
+                    }.bind(this)
+                }
+            });
+
             this.checkboxOnlyDirectChildren = new Ext.form.Checkbox({
                 name: "onlyDirectChildren",
                 style: "margin-bottom: 5px; margin-left: 5px",
@@ -296,6 +334,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                 this.languageInfo, "-",
                 this.toolbarFilterInfo,
                 this.clearFilterButton, "->",
+                this.selectObjectType, "-",
                 this.checkboxOnlyDirectChildren, "-",
                 this.exportButton, "-",
                 this.columnConfigButton,
--- a/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -247,7 +247,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                 xtype: "combo",
                 displayField:'name',
                 valueField: "value",
-                hidden: !this.element.data.general.allowInhreitance,
+                hidden: !this.element.data.general.allowInheritance,
                 store: selectObjectOptions,
                 editable: false,
                 width : 300,
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -591,18 +591,6 @@ class GridHelperService
             $conditionFilters[] = '(`path` = ' . $quotedPath . ' OR `path` like ' . $quotedWildcardPath . ')';
         }
 
-        if (isset($requestParams['all_objects']) && $requestParams['all_objects'] === 'true') {
-            $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
-        } else {
-            if (isset($requestParams['only_variant_objects']) && $requestParams['only_variant_objects'] === 'true') {
-                $conditionFilters[] = "`type` = 'variant'";
-            } elseif(isset($requestParams['only_objects']) && $requestParams['only_objects'] === 'true') {
-                $conditionFilters[] = "`type` = 'object'";
-            } else {
-                $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
-            }
-        }
-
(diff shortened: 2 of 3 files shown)


lucfumagalli:

lucfumagalli:
I just notice that when you try to filtering object via grid column the list doesn't update. Perhaps you have some suggestion or you can check better than me.

src/Helper/GridHelperService.php
```diff
@@ -591,6 +591,18 @@ public function prepareListingForGrid(array $requestParams, string $requestedLan
...
 
+        if (isset($requestParams['all_objects']) && $requestParams['all_objects'] === 'true') {
+            $conditionFilters[] = "`type` = 'object' OR `type` = 'variant'";
+        } else {
+            if (isset($requestParams['only_variant_objects']) && $requestParams['only_variant_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'variant'";
+            } elseif(isset($requestParams['only_objects']) && $requestParams['only_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'object'";
+            } else {
+                $conditionFilters[] = "`type` = 'object' OR `type` = 'variant'";
```
```suggestion
            $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
        } else {
            if (isset($requestParams['only_variant_objects']) && $requestParams['only_variant_objects'] === 'true') {
                $conditionFilters[] = "`type` = 'variant'";
            } elseif(isset($requestParams['only_objects']) && $requestParams['only_objects'] === 'true') {
                $conditionFilters[] = "`type` = 'object'";
            } else {
                $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
```
See also https://github.com/pimcore/admin-ui-classic-bundle/blob/fa2d03aaeb4787e1813f71f6a7ec58b28cd820cb/src/Helper/GridHelperService.php#L638

because otherwise every other condition might be "skipped" when `type` is `variant`.

lucfumagalli:
Now it works great!

src/Helper/GridHelperService.php
```diff
@@ -591,6 +591,18 @@ public function prepareListingForGrid(array $requestParams, string $requestedLan
...
 
+        if (isset($requestParams['all_objects']) && $requestParams['all_objects'] === 'true') {
+            $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
+        } else {
+            if (isset($requestParams['only_variant_objects']) && $requestParams['only_variant_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'variant'";
+            } elseif(isset($requestParams['only_objects']) && $requestParams['only_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'object'";
+            } else {
+                $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
```
kingjia90: just thinking maybe we should use the setter and the constants, eg.
https://github.com/pimcore/admin-ui-classic-bundle/blob/fa2d03aaeb4787e1813f71f6a7ec58b28cd820cb/src/Helper/GridHelperService.php#L673-L675
to be easier to read and mantain.

In there, we use a `WHERE IN` operator (instead of proposed OR), which also would work.

public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -231,6 +231,44 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
 
             this.store.getProxy().setExtraParam("query", this.searchFilter);
 
+            var selectObjectOptions = Ext.create('Ext.data.Store', {
```
kingjia90, line 234: It would be nice if this could be hidden when the class does NOT "Allow Variants" or so, just to avoid cluttered screen and also because there would be no any usefulness, eg. no any relevant change when switching between the select options.

translations/admin.en.yaml
```diff
@@ -432,6 +432,9 @@ system_columns: 'System columns'
 columns: Columns
 children_grid: 'Children Grid'
 only_children: 'just direct children'
+only_object: 'just objects'
+only_variant: 'just variants object'
+select_objects_type: "Object's type to show"
```
kingjia90:
```suggestion
select_objects_type: "Type to show"
```
being already in the context of objects, likely no need to specify it

src/Helper/GridHelperService.php
```diff
@@ -591,6 +591,18 @@ public function prepareListingForGrid(array $requestParams, string $requestedLan
             $conditionFilters[] = '(`path` = ' . $quotedPath . ' OR `path` like ' . $quotedWildcardPath . ')';
         }
 
+        if (isset($requestParams['all_objects']) && $requestParams['all_objects'] === 'true') {
+            $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
+        } else {
```
kingjia90: technically speaking this line could be avoided, just to keep the former behaviour and because for `all objects` we want to be more a "de-facto" all.

Overall this block of code has some redundancy, maybe should be shorten a bit but using a null coalescing assignment

kingjia90 (changes requested):

public/js/pimcore/object/helpers/gridTabAbstract.js
lucfumagalli: Now this should be setted! 👍🏼

src/Helper/GridHelperService.php
lucfumagalli: I added this few lines because the first time you load a folder you can correctly change the type of objects to see. But on the second change it stays stuck on the first one.

src/Helper/GridHelperService.php
lucfumagalli: Based on your suggestion, everything should work much better now. The code comes out even cleaner and with less redundancy. 

There was a problem with the javascript code, whenever i change the select option i setted a new parameter instead of modify an existing one.

src/Helper/GridHelperService.php
lucfumagalli: Reply on the first comment.

public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -231,6 +231,45 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
...
+            });
+
+            this.selectObjectType = new Ext.form.ComboBox({
+                fieldLabel: t('select_objects_type'),
+                name: 'objects_type',
+                labelWidth: 120,
+                xtype: "combo",
+                displayField:'name',
+                valueField: "value",
+                hidden: !this.element.data.general.allowInhreitance,
```
blankse:
```suggestion
                hidden: !this.element.data.general.allowInheritance,
```

public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -231,6 +231,45 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
 
             this.store.getProxy().setExtraParam("query", this.searchFilter);
 
+            var selectObjectOptions = Ext.create('Ext.data.Store', {
+                fields: ['name', 'value'],
+                data: [
+                    [t("empty"), "all_objects"],
```
kingjia90:
```suggestion
                    [t("all_types"), "all_objects"],
```
recommending to use this existing translation key that sounds slightly better than `empty`, at least in this context

src/Helper/GridHelperService.php
```diff
@@ -591,6 +591,18 @@ public function prepareListingForGrid(array $requestParams, string $requestedLan
...
+            if (isset($requestParams['only_variant_objects']) && $requestParams['only_variant_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'variant'";
+            } elseif(isset($requestParams['only_objects']) && $requestParams['only_objects'] === 'true') {
+                $conditionFilters[] = "`type` = 'object'";
+            } else {
+                $conditionFilters[] = "(`type` = 'object' OR `type` = 'variant')";
+            }
+        }
+
```
```suggestion
```
since we moved  the logic to `setObjectTypes`, this entire SQL condition is not needed anymore

src/Helper/GridHelperService.php
```diff
@@ -671,7 +683,13 @@ public function prepareListingForGrid(array $requestParams, string $requestedLan
...
 
         if ($class->getShowVariants()) {
-            $list->setObjectTypes([DataObject::OBJECT_TYPE_OBJECT, DataObject::OBJECT_TYPE_VARIANT]);
+            if (isset($requestParams['listing']) && $requestParams['listing'] == "only_objects") {
+                $list->setObjectTypes([DataObject::OBJECT_TYPE_OBJECT]); 
+            } elseif(isset($requestParams['listing']) && $requestParams['listing'] == "only_variant_objects") {
+                $list->setObjectTypes([DataObject::OBJECT_TYPE_VARIANT]); 
+            } else {
+                $list->setObjectTypes([DataObject::OBJECT_TYPE_OBJECT, DataObject::OBJECT_TYPE_VARIANT]);
+            }
```
kingjia90: This needs to be refactor more like

```
        if ($class->getAllowVariants()) {
            if ($class->getShowVariants()) {
                $list->setObjectTypes([DataObject::OBJECT_TYPE_OBJECT, DataObject::OBJECT_TYPE_VARIANT]);
            }
            if (isset($requestParams['listing'])){
                if ($requestParams['listing'] === "only_objects") {
                    $list->setObjectTypes([DataObject::OBJECT_TYPE_OBJECT]);
                } elseif($requestParams['listing'] === "only_variant_objects") {
                    $list->setObjectTypes([DataObject::OBJECT_TYPE_VARIANT]);
                }
            }
        }
```

kingjia90 (changes requested):
Thank you for applying the changes, i have some further inputs.
Other than that, the rest looks good to me and it would be ready to be merged.

src/Helper/GridHelperService.php
kingjia90: Extra: Not sure if `ShowVariants` is even in use or needed here, as i see is more related to the tree than the grid
image
but let's keep it for the moment.

public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -231,6 +231,45 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
...
+                width : 300,
+                triggerAction: 'all',
+                value: 'all_objects',
+                listeners: {
+                    change: function(comboBox,selected){
+                        this.grid.getStore().setRemoteFilter(false);
+                        this.grid.filters.clearFilters();
+                        this.grid.getStore().clearFilter();
+
+                        this.store.getProxy().setExtraParam("listing", selected);
```
kingjia90: not minding this, but it would be nicer to have something more "descriptive"
```suggestion
                        this.store.getProxy().setExtraParam("filter_by_object_type", selected);
```
in case this suggestion is applied, please also adapt all the former usages

src/Helper/GridHelperService.php
lucfumagalli: I replaced the code with your. It works great.

lucfumagalli:
All requests have been pushed. Are we now ready to merge? 
Thank you for your support.

kingjia90:
Yes, it's merged now


Grazie ancora per il tuo contributo e ti auguro un buon finesettimana 👍🏻

PR #466, opened 2024-03-19
get date with timezone offset

BlackbitDevs:
Date field doesn't respect timezone now.
This PR add timezone offset to date value and prevent saving date timestamp for day before.

When setting the date, the time is set to 00:00 and the timezone offset is subtracted and, as a result, the date is set to yesterday


markus-moser:
The timezone topic has some more problems and we decided to go for an alternative approach. Take a look at https://github.com/pimcore/pimcore/issues/16184#issuecomment-2145188141 for an overview. 

Your case should then be fixed too at the end with this PR: https://github.com/pimcore/admin-ui-classic-bundle/pull/553

markus-moser:
Thanks a lot for your input, it definitely helped figuring out the whole topic 👍

PR #467, opened 2024-03-21
[BUGFIX] Fixed dirty detection of carbon fields

turbo-ele:


src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1310,8 +1311,18 @@ public function saveAction(Request $request): JsonResponse
...
-                    $getter  ='get' . ucfirst($field);
-                    if ($object->$getter() !== $objectFromDatabase->$getter()) {
+                    $getter = 'get' . ucfirst($field);
+                    $objectValue = $object->$getter();
+                    $objectFromDatabaseValue = $objectFromDatabase->$getter();
+                    if ($objectValue !== $objectFromDatabaseValue) {
+                        // For Carbon objects, we need to compare the values - not the objects.
+                        if (
+                            $objectValue instanceof Carbon
+                            && $objectFromDatabaseValue instanceof Carbon
```
fashxp: I think we should rather go for the [`EqualComparisonInterface`](https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/ClassDefinition/Data/EqualComparisonInterface.php#L26) of the field definitions, wdyt?

fashxp (changes requested):

src/Controller/Admin/DataObject/DataObjectController.php
turbo-ele: Yes, you are completely right.
I adapted it - is that what you had in mind?

fashxp (approved):

fashxp (approved):

fashxp:
thx very much!

PR #468, opened 2024-03-21
Fixed an issue of save, publish and unpublish object when it is not allowed

rahul-pim-dev:
See: https://github.com/pimcore/pimcore/issues/15153

### DIFF
--- a/public/js/pimcore/object/object.js
+++ b/public/js/pimcore/object/object.js
@@ -851,6 +851,11 @@ pimcore.object.object = Class.create(pimcore.object.abstract, {
     },
 
     publish: function (only, callback) {
+
+        if(!this.isAllowed('publish')) {
+            return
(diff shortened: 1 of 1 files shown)


rahul-pim-dev:

rahul-pim-dev:
recheck

blankse:
Target branch of bug fixes should be `1.4`

fashxp:
this actually doesn't resolve https://github.com/pimcore/pimcore/issues/15153

and I don't think we should merge that change, as it just silently drops save commands and users don't get any feedback anymore that elements are not saved.

fashxp:
Since there is no further feedback, I'll close this for now. 
Feel free to continue the discussion or create a new PR if there are new findings or aspects. 

Thank you very much.

PR #469, opened 2024-03-21
[BUGFIX]: Moving objects in alphabetical tree

dturopoli:
### **Steps to reproduce**

- set "Sort children by" to Key (A-Z) (same thing for Z-A) to parent object/folder
- try to move child objects inside parent
- message appears that moving is not possible (this is correct), but after refresh of parent child is moved to nearest sibling to cursor while droping child


robertSt7:
@dturopoli Could you please rebase this to 1.4? Thanks

PR #471, opened 2024-03-22
Only run POEditor Export in main repository not in forks

matthiashamacher:


martineiber (approved):

PR #472, opened 2024-03-23
Add possibility to make languages not mandatory

### LINKED ISSUE #2958: Only one language mandatory
Hi,

For a client (for several projects even) need to make only one language mandatory on the set of localized fields. Assuming that my content is localized in 3 languages (FR, EN, ES), we would like only one of these 3 languages to be required during registration and publication.

Cheers,

matthiashamacher:
## Changes
Resolves pimcore/pimcore#2958

## Additional infos
This uses the logic introduced by pimcore/pimcore#16835 to make languages not mandatory. We add a checkbox in the System Setting where the user can set which languages should be mandatory. In the display logic of the localizedfield fields we update the mandatory state of the fields according to the languages mandatory need.

### DIFF
--- a/src/Controller/Admin/IndexController.php
+++ b/src/Controller/Admin/IndexController.php
@@ -206,6 +206,8 @@ class IndexController extends AdminAbstractController implements KernelResponseE
             $adminEntrypointUrl = null;
         }
 
+        $requiredLangauges = $systemSettings['general']['required_languages'];
+
         $settings = [
             'instanceId'          => $this->getInstanceId(),
             'version'             => Version::getVersion(),
@@ -224,7 +226,7 @@ class IndexController extends AdminAbstractController implements KernelResponseE
                 $systemSettings['general']['valid_languages'],
                 true
             ),
-            'requiredLanguages' => $systemSettings['general']['required_languages'],
+            'requiredLanguages' => empty($requiredLangauges) === true ? $systemSettings['general']['valid_languages'] : $requiredLangauges,
 
             // flags
             'showCloseConfirmation'          => true,
--- a/src/Controller/Admin/IndexController.php
+++ b/src/Controller/Admin/IndexController.php
@@ -206,7 +206,7 @@ class IndexController extends AdminAbstractController implements KernelResponseE
             $adminEntrypointUrl = null;
         }
 
-        $requiredLangauges = $systemSettings['general']['required_languages'];
+        $requiredLanguages = $systemSettings['general']['required_languages'];
 
         $settings = [
             'instanceId'          => $this->getInstanceId(),
@@ -226,9 +226,7 @@ class IndexController extends AdminAbstractController implements KernelResponseE
                 $systemSettings['general']['valid_languages'],
                 true
             ),
-            'requiredLanguages' => empty($requiredLangauges) === true
-                ? $systemSettings['general']['valid_languages']
-                : $requiredLangauges,
+            'requiredLanguages' => $requiredLanguages ?: $systemSettings['general']['valid_languages'],
 
             // flags
             'showCloseConfirmation'          => true,
--- a/src/Controller/Admin/IndexController.php
+++ b/src/Controller/Admin/IndexController.php
@@ -226,7 +226,9 @@ class IndexController extends AdminAbstractController implements KernelResponseE
                 $systemSettings['general']['valid_languages'],
                 true
             ),
-            'requiredLanguages' => empty($requiredLangauges) === true ? $systemSettings['general']['valid_languages'] : $requiredLangauges,
+            'requiredLanguages' => empty($requiredLangauges) === true
+                ? $systemSettings['general']['valid_languages']
+                : $requiredLangauges,
 
             // flags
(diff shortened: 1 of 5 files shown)


src/Controller/Admin/IndexController.php
```diff
@@ -224,6 +224,7 @@ protected function buildPimcoreSettings(
                 $systemSettings['general']['valid_languages'],
```
mcop1, line 229:
```suggestion
                $systemSettings['general']['valid_languages'] ?? [],
```
@matthiashamacher this could be null now, since we removed the default value in https://github.com/pimcore/pimcore/pull/16835 for it, right?

mcop1 (commented):
Additionally it seems to break the system settings ui now. Could you please have a look?

src/Controller/Admin/IndexController.php
matthiashamacher: Not null as its defined to be an array. So it would always be an empty array. But with the logic change from pimcore/pimcore#16835 an empty array does not make any sense as then we would have no required language so I overwrite an empty array with the valid_languages as in pimcore/pimcore#16835

matthiashamacher:
I cannot find any problems with the system settings ui. What exactly did break? I had an merge conflict in the translations, that is fixed now, maybe it was that?

src/Controller/Admin/IndexController.php
```diff
@@ -206,6 +206,8 @@ protected function buildPimcoreSettings(
             $adminEntrypointUrl = null;
         }
 
+        $requiredLangauges = $systemSettings['general']['required_languages'];
```
blankse: Typo in variable name
```suggestion
        $requiredLanguages = $systemSettings['general']['required_languages'];
```

src/Controller/Admin/IndexController.php
```diff
@@ -224,6 +226,9 @@ protected function buildPimcoreSettings(
                 $systemSettings['general']['valid_languages'],
                 true
             ),
+            'requiredLanguages' => empty($requiredLangauges) === true
+                ? $systemSettings['general']['valid_languages']
+                : $requiredLangauges,
```
blankse: empty() function is not necessary because variable always exists
```suggestion
            'requiredLanguages' => $requiredLanguages ?: $systemSettings['general']['valid_languages'],
```

src/Controller/Admin/IndexController.php
mcop1: Still technically the array key could be missing, when you didnt save the system settings with the new required languages features at least once:
image

mcop1:
> I cannot find any problems with the system settings ui. What exactly did break? I had an merge conflict in the translations, that is fixed now, maybe it was that?

It is working now, sorry, seems like it was a problem on my end of the side.

src/Controller/Admin/IndexController.php
matthiashamacher: I'm not quite sure how you got that error, I cannot reproduce it. I have removed the key from the `var/config/system_settings/system_settings.yaml` but still I cannot get that error.
Even though we remove the default value the key is still present in the config but as empty array. This is because of the prototype, which we have for the required_languages: https://github.com/pimcore/pimcore/pull/16835/files#diff-bab0da2e129502ead20a1f1ac979fd2260c5d4c8da202b6a8fbf1e9d906d4541R232
If you prototype an array node with scalar, then you get an empty array 
`->prototype('scalar')->end()` exists:
CleanShot 2024-04-20 at 10 53 04
`->prototype('scalar')->end()` removed:
CleanShot 2024-04-20 at 10 53 18

mcop1: Aaah, ok, the problem is, that I am using the `settings-store` as a `write_target` for my system settings, not the `symfony-config` option. 
Could you please try to change your `write_target` (https://pimcore.com/docs/platform/Pimcore/Tools_and_Features/System_Settings/) very briefly and see if you can reproduce the error?

matthiashamacher: Even then I cannot reproduce that error...
Change `write_target` location:
CleanShot 2024-04-22 at 08 54 48
Saved config in the `settings_store` table:
CleanShot 2024-04-22 at 08 55 06
Still getting no error and on accessing the configs all checkboxes are set as if there is an empty array returned.

mcop1: Can we please give it one more shot? Could you please try to set both `write_target` and `read_target` to settings-store? 
Sorry for the troubles, but I was able to reproduce the issue on another machine/env too.

matthiashamacher: You're right. If both are set to settings-store I also get that error. If implemented a check to prevent that. But in general I think there then is another bug regarding loading of settings, as in my opinion, if read_target is set to null it should use the same as write_target. And that doesn't seem to be the case currently.

mcop1: Thank you for adding the check. 
I agree, currently the fallback is always `symfony-config`, but it would make more sense to set it to the value of `write_target`.

PR #473, opened 2024-03-27
[Thumbnail] Mirror image via thumbnail configuration

youwe-petervanderwal:
Technically everything was prepared already in Pimcore since https://github.com/pimcore/pimcore/commit/6c47f7f9b34e779981815030e20f26f33dc300af#diff-6e26e398c3b9ab26394bfc4f6ef484880c9c1bd00d9b650ebd428dfa263e26f6

However, it seems that it never made it to the admin configuration.

Better late than never :)

### DIFF
--- a/public/js/pimcore/settings/thumbnail/item.js
+++ b/public/js/pimcore/settings/thumbnail/item.js
@@ -855,6 +855,38 @@ pimcore.settings.thumbnail.items = {
         return item;
     },
 
+    itemMirror: function (panel, data, getName) {
+        const niceName = t("mirror");
+        if (typeof getName != "undefined" && getName) {
+            return niceName;
+        }
+
+        if (typeof data == "undefined") {
+            data = {};
+        }
+        const myId = Ext.id();
+
+        return new Ext.form.FormPanel({
+            id: myId,
+            style: "margin-top: 10px",
+            border: true,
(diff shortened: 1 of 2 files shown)


robertSt7:
@youwe-petervanderwal All works well. One small thing, could you please replace `var` with `const` and `let`? Thanks :)

youwe-petervanderwal:
> @youwe-petervanderwal All works well. One small thing, could you please replace `var` with `const` and `let`? Thanks :)

@robertSt7 Sure, I copied it from the other definitions, do you like me to fix the complete file or only this new part?

robertSt7:
@youwe-petervanderwal Thanks for the fast response. Only the new part please :)

robertSt7:
@youwe-petervanderwal Thanks a lot. 
One more thing. It seems that you translated the values for nl.
We handle this kind of stuff a little bit different. We actually only update the English translations keys.
The keys are then synced to PO Editor. https://poeditor.com/join/project/XliCYYgILb
You can then translate it there to different languages.

We export those languages on a regular basis to all bundles.

The default fallback is english anyways.
Could you please remove all updates from the non-english files?

Thank you!

robertSt7:
@youwe-petervanderwal Thanks a lot for your contribution :)

PR #474, opened 2024-03-29
Load all quantity values in Baseunit store

### LINKED ISSUE #16874: [Bug]: Base Unit list is limited to 25 results in Quantity Value list
### Pimcore version

11.1.6

### Steps to reproduce

- Add more than 25 units in Quantity values (Settings -> Data Objects -> Quantity Value)
- Open Base unit list
image

- 25 results are shown instead all units

### Actual Behavior

Only 25 units in list are shown, instead a whole list.

### Expected Behavior

Whole list should be shown.

Maksonis:
Resolves: [#16874](https://github.com/pimcore/pimcore/issues/16874)

### DIFF
--- a/public/js/pimcore/object/quantityvalue/unitsettings.js
+++ b/public/js/pimcore/object/quantityvalue/unitsettings.js
@@ -84,6 +84,12 @@ pimcore.object.quantityValue.unitsettings = Class.create({
                 reader: {
                     type: 'json',
                     rootProperty: 'data'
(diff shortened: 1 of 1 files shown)


blankse:
The target branch of bug fixes should be `1.4`

Maksonis:
@blankse done: https://github.com/pimcore/admin-ui-classic-bundle/pull/475

PR #476, opened 2024-03-31
Revert part of PSDP-244/686e913 to fix an error when uploading a new avatar

skoch98:
By uploading a new avatar this happends:
image

But it should be:
image

See https://github.com/pimcore/admin-ui-classic-bundle/commit/686e913fae7523981eb8a9ed682d2e2089627917#diff-efc7c8f1d31756b699ed50c0214db7ca6f13bf9544340e02d33e64624a5d3b84R3231


blankse:
@skoch98 Target branch of bug fixes should be `1.4`

PR #480, opened 2024-04-03
Login: Center custom logo

blankse:
I would prefer to center the logo in the login. For logos with a width of less than 300px, it sticks to the left edge.


martineiber (approved):

PR #488, opened 2024-04-11
[Bug, EC]: Fix unchecked array access in WorkflowManagementListener

### LINKED ISSUE #487: [Bug, EC]: Unchecked array access in WorkflowManagementListener causes warning in backoffice when moving document
### Description

As the moving of elements requires the settings permission, which also shows the settings tab, I allowed settings in the workspace permissions, but forbid it on every workflow place. This hides the settings tab, but allows to move the element - in this case a document (page).
But in doing so, I get a warning in the backoffice when moving an item (see screenshot).

This is the relevant code and it can be found here: 

`$data['userPermissions'] = array_merge(
    (array)$data['userPermissions'],
    $placeConfig->getUserPermissions($workflow, $element)
);`

A possible solution would be to preset the array index if it is not set beforehand:

`if (!isset($data['userPermissions'])) {
     $data['userPermissions'] = [];
}`

### How to reproduce

- Create a backoffice user with all workspace permissions for documents except versions and properties
- Create a workflow for those documents and forbid settings for all places
- Create pages and try to move them

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/487

### DIFF
--- a/src/EventListener/WorkflowManagementListener.php
+++ b/src/EventListener/WorkflowManagementListener.php
@@ -103,7 +103,7 @@ class WorkflowManagementListener implements EventSubscriberInterface
             foreach ($this->workflowManager->getOrderedPlaceConfigs($workflow, $marking) as $placeConfig) {
                 if (!$permissionsRespected && !empty($placeConfig->getPermissions($workflow, $element))) {
                     $data['userPermissions'] = array_merge(
-                        (array)$data['userPermissions'],
+                        (array)$data['userPermissions'] ?? [],
                         $placeConfig->getUserPermissions($workflow, $element)
                     );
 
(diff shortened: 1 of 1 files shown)


src/EventListener/WorkflowManagementListener.php
```diff
@@ -103,7 +103,7 @@ public function onAdminElementGetPreSendData(GenericEvent $e): void
             foreach ($this->workflowManager->getOrderedPlaceConfigs($workflow, $marking) as $placeConfig) {
                 if (!$permissionsRespected && !empty($placeConfig->getPermissions($workflow, $element))) {
                     $data['userPermissions'] = array_merge(
-                        (array)$data['userPermissions'],
+                        (array)($data['userPermissions'] ?? []),
```
kingjia90: Is that too hard to read and redundant typecast when not set? should i switch to something more "classic"?
```suggestion
                        isset($data['userPermissions']) ? (array)$data['userPermissions'] : [],
```

PR #496, opened 2024-04-17
[Bug]: Fix "getSaveData is not a function" in Folder context

mindspace-devs:
When Editing Relations in a Folder overview, the beforeLoad listener tried to call this.object.getSaveData
But the Context of this.object in the Folder view is different than in a Object and the getSaveData function doesn't exist

This fixes the error and the rest of the editing works normally again.

### DIFF
--- a/.github/ci/files/.env
+++ b/.github/ci/files/.env
@@ -1,0 +1,2 @@
+APP_ENV=test
+APP_DEBUG=true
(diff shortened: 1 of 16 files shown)


mindspace-devs:

martineiber:
@mindspace-devs the target Branch for this bug fix should be 1.4! Could you change it to 1.4? Thx :)

PR #500, opened 2024-04-19
[Improvement] Changed @Route annotations with PHP attribute syntax

aashan10:
Used the `#[Route()]` syntax in place of the `/** @Route() **/` syntax


mattamon:
@aashan10 we will discuss internally if we still want to have attributes in classic-ui.

mattamon:
Hey @aashan10 we discussed it now internally and unfortunately we have to close this PR. 
We are currently working with full effort on the new studio-ui and want to keep the admin-ui as stable as possible.

Thanks for your contribution though! It is always welcome and really nice to see a familiar face in an PR! :)

Issue #497, opened 2024-04-17
[Bug]: Upload Dialog remains visible on failure. 


martineiber:
When using the `pimcore.helpers.uploadDialog()` https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/public/js/pimcore/helpers.js#L1019

In case of an error from the Backend, the upload Dialog says and can not be removed. 

The console shows following error: 
`Uncaught ReferenceError: pbar is not defined`

![image](https://github.com/pimcore/admin-ui-classic-bundle/assets/11687066/61b140ce-b019-4e79-9313-2da68fd308a0)

blankse:
@martineiber I created a PR: https://github.com/pimcore/admin-ui-classic-bundle/pull/501
Can you test it?

martineiber:
Thank you for the fix :)

Fixed by https://github.com/pimcore/admin-ui-classic-bundle/pull/501

PR #501, opened 2024-04-19
Fix: Upload Dialog remains visible on failure

blankse:
Fix: https://github.com/pimcore/admin-ui-classic-bundle/issues/497
Regression from https://github.com/pimcore/admin-ui-classic-bundle/pull/137

### DIFF
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -1078,10 +1078,10 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                 win.show();
 
                 let finishedErrorHandler = function (e) {
-                    this.activeUploads--;
+                    activeUploads--;
                     win.remove(pbar);
 
-                    if(this.activeUploads < 1) {
+                    if (activeUploads < 1) {
                         win.close();
                     }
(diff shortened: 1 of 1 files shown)


public/js/pimcore/helpers.js
```diff
@@ -1077,10 +1078,10 @@ pimcore.helpers.uploadDialog = function (url, filename, success, failure, descri
                 win.show();
 
                 let finishedErrorHandler = function (e) {
-                    this.activeUploads--;
+                    activeUploads--;
```
fashxp, line 1083: do we really need a global global variable here? Or do I miss something?

fashxp (dismissed):

public/js/pimcore/helpers.js
blankse: @fashxp The variable is defined here: https://github.com/pimcore/admin-ui-classic-bundle/pull/501/files#diff-efc7c8f1d31756b699ed50c0214db7ca6f13bf9544340e02d33e64624a5d3b84R1089

Old line 1088, New line 1089

martineiber (approved):

PR #503, opened 2024-04-19
Fix null to property error in operator text

saschabaecher-twocream:
## Situation:
- Open Grid View from folder
- In the Grid View configuration, select the "text" operator under "Formatters".
- Now we get the following error:
grafik

## Expected behavior:
That it is possible to use the operator text without getting an error message for which the user is not responsible.

## Workaround:
We simply set the defaultText as the default value for the label instead of `null`.


fashxp:
thx very much

PR #507, opened 2024-04-22
Fix: Numeric Editable config option "decimalPrecision" does not work

blankse:
Fix: https://github.com/pimcore/pimcore/issues/16959


lukmzig:
@blankse thank you very much for the fix

PR #508, opened 2024-04-23
NotificationController: Set limit/offset as integer

blankse:
See: https://github.com/pimcore/pimcore/pull/16968


robertSt7:
@blankse Thanks a lot :)

PR #379, opened 2023-12-28
[Bugfix] textLayoutPreviewAction does not assign a name to the text element

dark-cms:
To be able to use the preview in the class editor, an object with a name is expected. 
The solution provided is only used for the preview and has no effect on other areas.

Currently, the preview options in the demo systems of Pimcore do not seem to work and throw an error 500

PR #518, opened 2024-04-30
[FIX]: Dynamic Text Labels preview

Navino16:
When going to the preview of a Dynamic Text Labels we get the following error:
image

After fixing this issue, when we use a Custom Renderer Class we get this error:
image

This PR fix this 2 errors.

Also, the first fix is a duplicate of #379 which seems to be inactive


robertSt7:
@Navino16 Thanks for the fix

PR #519, opened 2024-04-30
[Asset upload] Provide translation parameters in correct order

BlackbitDevs:
As written in https://github.com/pimcore/pimcore/pull/17005:
Translation for `asset_type_change_not_allowed` for `en` is for example 
> Only assets of same type are allowed here.</strong><br/>[ type of uploaded asset: "%s" | type of existing asset: "%s" ]

but in https://github.com/pimcore/admin-ui-classic-bundle/blob/5536216395824be91fda28884c911555de6bf28a/src/Controller/Admin/Asset/AssetController.php#L579 the parameters are provided in reverse order.


alexz707:
Hello @BlackbitDevs
thanks a lot for finding and fixing this bug!
PR LGTM :)

PR #524, opened 2024-05-08
[Improvement, EC]: Display read-only asset metadata when granted view permission

### LINKED ISSUE #435: [Improvement, EC]: Display read-only asset metadata when granted view permission
### Pimcore version

11.x

### Actual Behavior

In the backoffice, the asset metadata tab disappears when either the workflow state disallows editing (modify, save, publish, unpublish, delete, rename false) or a user is lacking publish rights, so the meta data can not be read.
The metadata is only visible when it is editable.

### Expected Behavior

The metadata should be visible by correct rights.

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/435
Requires https://github.com/pimcore/asset-metadata-class-definitions/pull/186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+#### v1.5.0
+- [Assets] Metadata can be now displayed as a read-only tab when the user is granted `view` permissions to the asset.
+  
 #### v1.4.0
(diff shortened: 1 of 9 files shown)


markus-moser (approved):
I tested it and it works fine - good job 👍 

I'm still testing the asset metadata class definitions part

PR #527, opened 2024-05-10
Fix: Add video service providers to CSP allow list

### LINKED ISSUE #526: [Bug]: [Objects] External video preview image (YouTube) blocked due to CSP
### Pimcore version

11.2.2

### Steps to reproduce

1. Have data field of type `video`.
2. Choose external video (e.g. YouTube) and paste its YT-code.
3. Preview image is blocked by the browser due to CSP.

`Refused to frame 'https://www.youtube-nocookie.com/' because it violates the following Content Security Policy directive: "frame-src 'self' data:".`

### Actual Behavior

Preview image is blocked due to CSP and thus not shown.

### Expected Behavior

As configuring YouTube in video field is a core functionality it should not require project specific configuring such as:
```yaml
pimcore_admin:
    admin_csp_header:
        additional_urls:
            frame-src:
                - 'https://www.youtube-nocookie.com/'
```

Instead these external video services should be allowed by default in `ContentSecurityPolicyHandler`.

podarcis:
This adds the URLs of currently used video services to the CSP allow list for type frame so that video preview thumbnails are shown in backend.

Before this fix the video preview thumbnails are not shown but instead following message is logged to console:
`Refused to frame 'https://www.youtube-nocookie.com/' because it violates the following Content Security Policy directive: "frame-src 'self' data:".`

Fixes pimcore/admin-ui-classic-bundle#526

### DIFF
--- a/src/Security/ContentSecurityPolicyHandler.php
+++ b/src/Security/ContentSecurityPolicyHandler.php
@@ -56,6 +56,11 @@ class ContentSecurityPolicyHandler implements LoggerAwareInterface
         self::SCRIPT_OPT => [
             'https://buttons.github.io/buttons.js', // GitHub star button on login page
         ],
+        self::FRAME_OPT => [
+            'https://www.youtube-nocookie.com/', // Video preview thumbnail for YouTube
+            'https://www.dailymotion.com/',      // Video preview thumbnail for Dailymotion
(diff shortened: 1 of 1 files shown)


robertSt7:
@podarcis The target branch for bug fixes should be 1.4. Could you please change that? Thanks

robertSt7:
@podarcis Thanks for the fix :)

PR #530, opened 2024-05-16
[Bug]: Prevent displaying delete button on not allowed to delete buttons

kingjia90:
Resolves https://github.com/orgs/pimcore/discussions/17049


public/js/pimcore/object/folder/search.js
```diff
@@ -458,7 +458,7 @@ pimcore.object.search = Class.create(pimcore.object.helpers.gridTabAbstract, {
             }
 
             menu.add(new Ext.menu.Item({
-                hidden: data.data.locked,
```
cancan101, line 461: do you need this change down below (L487)  in the bulk change case?

kingjia90: Thank you for reviewing, yep, indeed we need a change there as well, but it could be tricky, for example, when selecting rows/element that are allowed to be deleted along the ones that are not allowed. 
Would need some more work, i have an idea and will try it out, will push the change soon

kingjia90: Something like https://github.com/pimcore/admin-ui-classic-bundle/pull/530/commits/925245f2b24e54947e7ce2eabf25dbf256ae7e9a would work, i will refine and clean it a little bit tomorrow, but would highly appreciate if you could review and give it a try if you have chance in the meanwhile

kingjia90: Possible drawbacks are that the $pathString list could be too long (occupying lots of the screen height) if one selects hundreds of rows to delete and only a few are not allowed to be deleted, in that case it would be ideal to display the rows that would be left untouched instead (if it would be shorter) and that there are 2 confirm popup in a row (the first to warn and the second which is the former one)

cancan101:
You might also need a similar fix to prevent bulk editing in the grid view. I have editing disabled for a given workflow state but am still able to edit the Object through the grid view.

mattamon:
Obsolete (in a nice way)

PR #531, opened 2024-05-16
Fixes problem when post_max_size is set to unlimited but upload_max_filesize has a limit

DanielLaranjeira063:
Currently, if you have in your server, lets say 100M limit for file uploads, but post_max_size is set to unlimited (by setting it's value to 0), the bundle will set the upload_max_filesize setting to 0 as it's taking the minimum value between both into account. and it won't let you upload any files as javascript is comparing the size of the file to the value of 0.
This PR changes that so that if the minimum value between post_max_size is set to 0, then it should only take max_upload_filesize into account. If that value is set to 0 as well, then it is correct, as max_upload_filesize does not have an unlimiter.
If none of them are set to 0, then it will take the minimum value between the two as it's supposed to.

### DIFF
--- a/.github/ci/files/.env
+++ b/.github/ci/files/.env
@@ -1,0 +1,2 @@
+APP_ENV=test
+APP_DEBUG=true
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,34 +1,14 @@
-name: "CLA Assistant"
+name: CLA check
+
 on:
   issue_comment:
     types: [created]
   pull_request_target:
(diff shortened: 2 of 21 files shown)


DanielLaranjeira063:

DanielLaranjeira063:
recheck

robertSt7:
@DanielLaranjeira063 The target branch for bug fixes should be 1.4. Could you please change the branch? Thanks

DanielLaranjeira063:
I will be closing this PR and open a new one then for 1.4 branch

PR #535, opened 2024-05-19
Add optimized admin loading to reverse relations

sl0wlydeadly:
Add "Enable Async Load in Admin" in reverseObjectRelation field

Pimcore Issue: https://github.com/pimcore/admin-ui-classic-bundle/issues/620

### DIFF
--- a/public/js/pimcore/object/tags/reverseObjectRelation.js
+++ b/public/js/pimcore/object/tags/reverseObjectRelation.js
@@ -178,7 +178,7 @@ pimcore.object.tags.reverseObjectRelation = Class.create(pimcore.object.tags.man
                 markDirty: false,
                 listeners: {
                     afterrender: function (gridview) {
-                        this.requestNicePathData(this.store.data);
+                        this.requestNicePathData(this.store.data, true);
                     }.bind(this)
                 }
             },
--- a/public/js/pimcore/object/classes/data/reverseObjectRelation.js
+++ b/public/js/pimcore/object/classes/data/reverseObjectRelation.js
@@ -209,6 +209,20 @@ pimcore.object.classes.data.reverseObjectRelation = Class.create(pimcore.object.
                 value: t('non_owner_description'),
                 cls: "pimcore_extra_label_bottom"
             }));
+
+            if(this.context == 'class') {
+                this.specificPanel.add({
+                    xtype: "checkbox",
+                    boxLabel: t("enable_admin_async_load"),
+                    name: "optimizedAdminLoading",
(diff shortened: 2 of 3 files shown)


fashxp:
Thx for the PR.

But somehow it doesn't work for me. When activating the async loading, there is no extra request fired when opening a data object.

data object with reverse many-to-many relation (with activated async loading):
image

compare with object with many-to-many relation (with activated async loading):
image

sl0wlydeadly:
Hello and thank you for the response.
I have tried it locally and it worked also for Reverse Object Relation i was able to see the get nice path request.
Is it possible that you tried it without the second change?
image
This is the one change which is supposed to enable the get nice path request for the reverse object relation.

Also there is another PR in pimcore/pimcore which is relevant for the transmission of the data (classname) for the reverse relations to the admin ui:
https://github.com/pimcore/admin-ui-classic-bundle/issues/620
It is preferable that you test those changes together for the full feature.
If there is something I havent taken into consideration or a change that I should make please notify me and I will look into it :) :)

fashxp:
I had a closer look at this ... 
I think it can't work, as there are no changes in the `getLayoutEdit` function of `tags/reverseObjectRelation.js` .... which is responsible for rendering the grid ... and thus also loading the nice paths. 

You just changed the `getLayoutShow` function of `tags/manyToManyObjectRelation.js` ... which is just used when field is in read only mode.

fashxp:
@sl0wlydeadly :ping_pong:

sl0wlydeadly:
@fashxp Hello and thank you for the reply. I will take a closer look at it this weekend. I will probably have to add a check also for the getLayoutEdit, nevertheless i will also test it to see if my changes will be effective.

public/js/pimcore/object/tags/reverseObjectRelation.js
```diff
@@ -178,7 +178,7 @@ pimcore.object.tags.reverseObjectRelation = Class.create(pimcore.object.tags.man
                 markDirty: false,
                 listeners: {
                     afterrender: function (gridview) {
-                        this.requestNicePathData(this.store.data);
+                        this.requestNicePathData(this.store.data, true);
```
sl0wlydeadly, line 181: After testing this, it worked loading the nice path also in getLayoutEdit (reload admin mode) and I did not encounter any disadvantages.

fashxp: :+1:  now it looks good. thx very much!

fashxp (approved):

Issue #434, opened 2024-02-20
[Bug]: Quantity Value Permissions


storbahn:
Is it necessary to check the permission for the quantity values in objects (request /admin/quantity-value/convert-all when opening the objects)? 

In our use case the editor/user should only have the read access to the objects/system but should't have write access to the objects/system. But at the moment we have to give him the permission to the QuantityValue Unit Defintions to open the objects without an error.

In my opinion it makes only sense to check the permission for the administration of the QuantityValue Unit Defintions
(Settings >> Data Objects >> Quantity Value) and not when opening an object.

PR #536, opened 2024-05-21
[BUG] remove checks on quantity value convert actions

Melsa-Huebel:
### Resolves: #434

Remove checkPermission method on specific rights for quantity value convert endpoints.

### Additional info
It seems not necessary to provide specific checks while using convert functions for quanity values. Futhermore if you giving the user the quantityValueUnits access right the user can edit all scale units. That seems not correct at least in our use case where just users with higher level access rights should be able to edit scale units.

### DIFF
--- a/src/Controller/Admin/DataObject/QuantityValueController.php
+++ b/src/Controller/Admin/DataObject/QuantityValueController.php
@@ -236,8 +236,6 @@ class QuantityValueController extends AdminAbstractController
      */
     public function convertAction(Request $request, UnitConversionService $conversionService): JsonResponse
     {
-        $this->checkPermission('quantityValueUnits');
-
         $fromUnitId = $request->get('fromUnit');
         $toUnitId = $request->get('toUnit');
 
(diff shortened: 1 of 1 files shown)


Melsa-Huebel:

robertSt7:
Hi @aMH-techsigns I think we should check the `objects` permission here. What is your opinion?

Melsa-Huebel:
> Hi @aMH-techsigns I think we should check the `objects` permission here. What is your opinion?

Hi @robertSt7 you mean to check the user permission for the underlaying data object tree, right? If so my thinking is if this is not handeled somewhere else maybe some steps before this call can even happen? So we might do stuff we shouldnt do on this position (might be unnecessary)? As the controller is inheriting from the admin controller where a admin user should be guaranteed i feel some kind of safe :-D.
Furthermore these two methods don't do any harm or update on the underlying data as far as i can see.

But ok i got you. You mean there is missing some kind of check (security-wise).
I can add the check on objects on both methods. It should not harm the process as the object permission needs to be granted to come this far.

Melsa-Huebel:
@robertSt7 I followed your suggestion and added the two permission checks. Thx

robertSt7:
@aMH-techsigns Thanks a lot for the fix. There is still a problem when the quantity value is changed the auto-save will be triggered and then it shows another error. I have alread created a follow- up for this. https://github.com/pimcore/admin-ui-classic-bundle/issues/544

PR #537, opened 2024-05-23
Update minipaint

robertSt7:


templates/admin/asset/image_editor.html.twig
```diff
@@ -127,13 +142,47 @@
...
+        })
+
+        removeHref();
+
+        document.getElementsByTagName('nav')[0].addEventListener('click', (e) => {
+            removeHref();
+        });
+    }
+
+    function removeHref(className = 'main_menu') {
```
markus-moser: the `className` parameter is unused -> let's remove it.

templates/admin/asset/image_editor.html.twig
```diff
@@ -127,13 +142,47 @@
...
+        const allLinkElements = document.getElementsByTagName('nav')[0].querySelectorAll("a");
+        for (let a of allLinkElements) {
+            if (a.getAttribute('href') === 'javascript:void(0)') {
+                a.removeAttribute('href');
```
markus-moser, line 187: Let's add a line like this here:

`a.classList.add('cursor-pointer');`

and some CSS

```CSS
.cursor-pointer {
    cursor: pointer;
}
```
As we removed the href the styling of the mouse cursor is now a bit broken (default mouse cursor instead of pointer).

templates/admin/asset/image_editor.html.twig
```diff
@@ -127,13 +142,47 @@
...
+            }
+        }
+    }
     
     window.addEventListener("load", function(e) {
```
markus-moser, line 192: Do we need this load event listener at all?

markus-moser (changes requested):

templates/admin/asset/image_editor.html.twig
robertSt7: Yes we call there the function loadEditor

markus-moser: Okay, but the `.then()` part with the `console.log`?

robertSt7: I think we can remove that

templates/admin/asset/image_editor.html.twig
markus-moser: Thanks, it works - we need the `.cursor-pointer` class also on the save button :)

markus-moser (approved):
Thanks a lot, now it's fine!

Good job 👍

PR #542, opened 2024-05-29
Decouple simple backend search from admin UI classic bundle

### LINKED ISSUE #124: Remaining hardcoded dependency to SimpleBackendSearchBundle
Fulltextsearch in grid still uses subquery to `search_backend_data` table. 

see https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/src/Helper/GridHelperService.php#L613

markus-moser:
Depends on https://github.com/pimcore/pimcore/pull/17105

Resolves #124:

- Search field is hidden when no search implementation is available
- Condition related to `search_backend_data` table is added via event listener instead of adding it directly


kingjia90 (approved):

public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -329,18 +298,56 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
...
+            this.clearFilterButton, "->",
+            this.selectObjectType, "-",
+            this.checkboxOnlyDirectChildren, "-",
+            this.exportButton, "-",
+            this.columnConfigButton,
+            this.saveColumnConfigButton
+        ];
+
+        if(pimcore.helpers.hasSearchImplementation()) {
+            this.searchField = new Ext.form.TextField(
```
kingjia90: LGTM, just leaving a note that we have also to pick the changes in this js file and apply to v1.4 as bug fix.
Otherwise, with the combination of pimcore 11.2 without Simple Search bundle installed would still show this search field and crash due to search_backend_data table being not existing when typing something to search

kingjia90: Resolved: I've created a PR and merged in 1.4, then upmerged 1.4 into 1.x and merged it in this branch
https://github.com/pimcore/admin-ui-classic-bundle/pull/546

markus-moser: Great, thanks for checking.

PR #545, opened 2024-05-31
FieldDefinition: Add missing forbidden names

blankse:
`Fatal error: Declaration of Pimcore\Model\DataObject\Test::setClassId(?string $classId) must be compatible with Pimcore\Model\DataObject\Concrete::setClassId($o_classId) in /var/www/html/var/classes/DataObject/Test.php on line 188`

See also https://github.com/pimcore/pimcore/pull/17109

### DIFF
--- a/public/js/pimcore/object/classes/data/data.js
+++ b/public/js/pimcore/object/classes/data/data.js
@@ -19,15 +19,16 @@ pimcore.object.classes.data.data = Class.create({
 
     invalidFieldNames: false,
     forbiddenNames: [
-        "apipluginbroker", "byid", "bypath", "cachetag", "cachetags", "childamount", "children", "childrensortby",
-        "childrensortorder", "childs", "class", "classid", "classname", "classtitle", "closestparentofclass",
-        "creationdate", "dao", "data", "definition", "dependencies", "dirtyfields", "dirtylanguages", "fieldname",
-        "fullpath", "haschildren", "hassiblings", "id", "idpath", "index", "key", "language", "latestversion",
-        "lazyloadedfieldnames", "list", "localizedfields", "locked", "modificationdate", "nextparentforinheritance",
-        "object", "omitmandatorycheck", "parent", "parentclass", "parentid", "path", "permissions",
-        "permissionsforuser", "properties", "property", "published", "relationdata", "resource", "scheduledtasks",
-        "siblings", "type", "usermodification", "userowner", "userpermissions", "value", "valueforfieldname",
-        "valuefromparent", "values", "versioncount", "versions",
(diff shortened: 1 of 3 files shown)


public/js/pimcore/object/classes/data/data.js
```diff
@@ -25,7 +25,7 @@ pimcore.object.classes.data.data = Class.create({
         "versions", "properties", "permissions", "permissionsforuser", "childamount", "apipluginbroker", "resource",
         "parentClass", "definition", "locked", "language", "omitmandatorycheck", "idpath", "object", "fieldname",
         "property", "localizedfields", "parentid", "scheduledtasks", "latestVersion", "haschildren", "siblings", "hassiblings",
-        "childrenSortby", "childrensortorder", "versioncount", "dirtylanguages", "dirtyfields", "classtitle"
+        "childrensortby", "childrensortorder", "versioncount", "dirtylanguages", "dirtyfields", "classtitle", "classid"
```
kingjia90: Nice catch, LGTM but i've noticed that it breaks for every other getter as well eg. `valueForFieldName`, `dao`, `NextParentForInheritance` ,`ClosestParentOfClass`, `LazyLoadedFieldNames`, `RelationData`

Could you please add these as well? TIA

image
image
image

blankse: @kingjia90 I added it. I also sorted it alphabetically. So we see faster missing properties.

kingjia90: Thank you for the changes and sorting, that's helpful.

Indeed, it's a little hard to read and manage such a long array, and we can't break a line for each element.
I am having a sort of brain-fart, what do you think about a "counter-intuitive" (just for human/dev readability) solution to make it an associative array, group by similar domain/areas and that gets flattened for the compiler? 🙈

blankse: @kingjia90 I don't think I quite follow you. Do you mean to make a multidimensional list out of it?

Maby we generate the list via Ajax call? Via PHP reflection?

We can write it down as an issue. But I think it's too much for this PR.

kingjia90: Yes, some kind of that, like
```
someconst: {
  'parent': ["parent", "parentclass", "parentid"],
  'id': ["id", "byid"]
}
```
and flatten/reduced it to be like `"parent", "parentclass", "parentid","id", "byid"`

but agree, that's not worth it to invest so much time on that.

So far LGTM as is, i will have a deeper review and merge soon or latest on Monday. Wish you a nice weekend!

blankse: @kingjia90 I wish you also a nice weekend!

robertSt7:
@blankse Thanks a lot

PR #547, opened 2024-06-03
Many-to-many object relation / multiselect in tag mode: support drag & drop to change order of assigned items

BlackbitDevs:
Assume we have a many-to-many object relation in display mode = tag field.

When assigning new objects, everything works great but sometimes the order of the assigned items matters and you want to insert an item in between or change the order of already assigned items.
So if you have assigned objects `/one` and `/three` and now want to insert `/two` in between, then you have to currently
1. delete assignment for`/three` 
2. assign `/two`
3. assign `/three`

This is quite complicated. And depending on number of assigned elements, this is can be even more work.

This PR adds drag & drop support for the assigned items. It works like in https://fiddle.sencha.com/fiddle/i65/preview
With this you can change the order of assigned objects afterwards.

The same functionality has also been applied to multiselect fields in tag mode.

### DIFF
--- a/public/js/lib/ext-plugins/DragDropTag.js
+++ b/public/js/lib/ext-plugins/DragDropTag.js
@@ -1,0 +1,118 @@
+/**
+ * @author Adam Borowski
+ * @see https://fiddle.sencha.com/#fiddle/i65
+ */
+Ext.define('cas.helper.plugin.DragDropTag', {
+    extend: 'Ext.plugin.Abstract',
+    alias: 'plugin.dragdroptag',
+    requires: [],
+    statics: {},
+    init: function (cmp) {
+        cmp.on('render', this.afterRender, this, { single: true });
+    },
+    afterRender: function () {
+        var me = this.getCmp();
+        me.boundList = me.getPicker();
+        me.dragGroup = me.dropGroup = 'MultiselectDD-' + Ext.id();
+        me.dragZone = Ext.create('Ext.dd.DragZone', me.itemList, {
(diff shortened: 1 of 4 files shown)


robertSt7:
Hi @BlackbitDevs could you please replace `var` with `let` and `const`? Thanks

BlackbitDevs:
@robertSt7 Done.

fashxp:
We don't want to be picky, but we need to be careful with compliance and copyrights.
 
So, I'm sorry, we cannot merge this PR, as it includes code from a foreign author that has not signed our CLA. 
Options to continue: 
- rewrite that piece of code that it is your own work
- get permission of the original author

BlackbitDevs:
@adamborowski is the code in https://fiddle.sencha.com/fiddle/i65/preview from you? Would it be ok to use it in Pimcore?

adamborowski:
> @adamborowski is the code in https://fiddle.sencha.com/fiddle/i65/preview from you? Would it be ok to use it in Pimcore?

Sure

BlackbitDevs:
@fashxp Is [above comment](https://github.com/pimcore/admin-ui-classic-bundle/pull/547#issuecomment-2324415498) enough?

fashxp:
yes it is, thx very much @adamborowski. maybe we can translate the comments to english, test once more then it is ready to merge.

BlackbitDevs:
Have translated the comments.

mcop1:
Thanks, looks good to me.

PR #549, opened 2024-06-04
[Improvement]: Make Dependencies tab hideable via config

### LINKED ISSUE #15426: [Improvement]: Resolve Dependencies via Messenger
### Improvement description

Ideally there is a way how we can make this feature optional.

kingjia90:
Partly resolves https://github.com/pimcore/pimcore/issues/15426#issuecomment-2089991720
Requires https://github.com/pimcore/pimcore/pull/17121


markus-moser:
For the web2print documents we would need to do the check for the dependencies tab too:

https://github.com/pimcore/web-to-print-bundle/issues/78

markus-moser (approved):

PR #550, opened 2024-06-04
Improve logic for classification store in diff_versions in data objects

melissakittl:
### Changes in this pull request
Improve logic for classification store in diff_versions. Changes logic from "and" to "or", and check if the group is only active in one version.
It should be possible to get a difference between two versions, even if a classification store group was added / removed in one version and is not present in den previous version.
image

### How to reproduce
1. Create Class with Classification Store
2. create object and save
3. Add classfication store group to object, add content to some fields and save
4. select the two versions for comparison
5. the new added classification store group is now visible as change (before nothing was shown)


robertSt7:
@melissakittl Thanks for the contribution

PR #551, opened 2024-06-05
Replace Flysystem for grid-export with native filesystem operations to reduce IOPS

turbo-ele:
## Summary
This merge request replaces Flysystem with native filesystem operations to address the issue of excessively high IOPS during export processes, especially when exporting big datasets. 

## Background
During routine monitoring, we identified that the CSV export functionality was generating an unexpectedly high IOPS, approximately 4.5 TB, for a relatively modest file size of 255 MB. This level of IOPS was significantly impacting system performance and increasing operational costs (~200$ on AWS EFS).

## Investigation
Detailed analysis revealed that the use of Flysystem for handling file operations in the export process was contributing to the high IOPS. Flysystem, while providing a convenient abstraction for file handling across different storage systems, was not efficient for our use case involving large data manipulations and frequent file writes, as for appending new data, the whole file was first read, extended and then fully written back.

## Changes Made
- **Direct File Operations**: We have switched to using PHP's native file handling functions (`fopen`, `fwrite`, `fclose`) to be able to use the possibility of appending data to the temporary export file instead of overwriting it over and over again.

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/src/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1283,6 +1283,7 @@ class DataObjectHelperController extends AdminAbstractController
 
         // We don't use flysystem here because we need to append to the file
         $stream = null;
+
         try {
             $newData = [];
             $csvFile = $this->getCsvFile($fileHandle);
--- a/src/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/src/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -18,6 +18,7 @@ namespace Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject;
 
 use League\Flysystem\FilesystemException;
 use League\Flysystem\UnableToReadFile;
+use League\Flysystem\UnableToWriteFile;
 use Pimcore\Bundle\AdminBundle\Controller\AdminAbstractController;
 use Pimcore\Bundle\AdminBundle\Event\AdminEvents;
 use Pimcore\Bundle\AdminBundle\Helper\GridHelperService;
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -1299,32 +1297,53 @@ public function doExportAction(
...
                     $line = implode($delimiter, $line);
-                    fwrite($temp, $line);
+                    $newData[] = $line;
+
                 } else {
-                    fwrite($temp, implode($delimiter, array_map([$this, 'encodeFunc'], $line)));
-                }
-                if ($i < $lineCount - 1) {
-                    fwrite($temp, "\r\n");
+                    $newData[] = implode($delimiter, array_map([$this, 'encodeFunc'], $line));
```
jdreesen, line 1312:
```suggestion
                    $newData[] = implode($delimiter, array_map($this->encodeFunc(...), $line));
```

dpfaffenbauer:
we run into rate limit issues on gcs with the export. But your solution does not seem to work properly with multiple instances right? you always write to the local temp file and never back to the storage.

turbo-ele:
@dpfaffenbauer that's correct, it won't work with multiple instances unless the configured temp folder is a shared volume.

dpfaffenbauer:
@turbo-ele not a viable option IMO.

turbo-ele:
@dpfaffenbauer yes you are right, it's not the most versatile approach as it doesn't work with multiple instances if the temp folder is not shared. But the currently implemented solution also isn't a suitable option in our case, as for the aforementioned reasons.
Since I don't know any other option, it might make sense to make it configurable, if the export should be done natively on the file system or via Flysystem and also which adapter to use.

Thanks for pointing out the issue with multiple instances.

dpfaffenbauer:
you can configure your temp folder to be on a different storage like a local tmp folder.

turbo-ele:
Yes, I know, but Flysystem is still reading and writing the whole file all the time. This is why I initially created this MR.
But while you might want to have flysystem to use s3 for the temp folder, you still might want to have the export somewhere different.

dpfaffenbauer:
distributed systems are tough ;)

mcop1:
Hello @turbo-ele ,

we really appreciate the work you put into the pull request, but after discussing it internally we came to the conclusion that we wont merge it right now. We are currently working on a new API over here https://github.com/pimcore/studio-backend-bundle, where we are already doing things like this in a different way. 
Of course you are welcome to head over there and provide feedback if you have any. 

Thank you for your efforts!

PR #554, opened 2024-06-06
[Improvement]: Set default date time formats on startup based on user locale

kingjia90:
Related https://github.com/pimcore/pimcore/pull/17137

Draft: check with the timezone related changes that are conflicting
Naming could be improved, something else instead of short/medium?


public/js/pimcore/helpers.js
```diff
@@ -3534,3 +3534,59 @@ pimcore.helpers.sendRequest = function (
...
+            format+= separator + 'H:i:s';
+        }
+        if (format === ''){
+            format = 'Y-m-d H:i:s';
+        }
+        return Ext.Date.format(value, format);
+    }
+}
+
+pimcore.helpers.intlDateFormatFromLocale = function (fallback){
```
kingjia90: this one doesn't work really well 😞 leading zero is missed

image
image

kingjia90: potentially solved by https://github.com/pimcore/admin-ui-classic-bundle/pull/554/commits/dfe9f77d6b03ef7721292423ff8e3793bffc17a2 but not sure if it's a good approach or good trade off to avoid using any extra external library and to avoid to save hardcoded formats in some multi dimensional array in the js

kingjia90: For time https://github.com/pimcore/admin-ui-classic-bundle/pull/554/commits/c5648d84c3a1d359df8103969b410eb9696e627d along https://github.com/pimcore/pimcore/pull/17137/commits/5e2a48f81a46cc3354ee498a507c7e20c06d3196
image

kingjia90:
Workish but i feel it's too much on the fly.

Should be using the approach of ext-locale eg. 
https://github.com/pimcore/admin-ui-classic-bundle/blob/1976b8f869ad450e1b3d2b9e4aebcc47f3487e7e/public/js/lib/ext-locale/locale-en_GB.js
by doing it once globally by eventually parsing the output of `Intl.DateTimeFormat` or is it bad to rely/depending on browser that [may be inconsistent](https://stackoverflow.com/questions/71937638/intl-datetimeformat-inconsistent-between-chrome-node-and-firefox)?

markus-moser:
@kingjia90 We should add the "Date Time Locale" setting also to the profile page:

image

image

markus-moser:
The date range data type should be adjusted too:

image

markus-moser:
System date columns are empty

image

public/js/pimcore/settings/user/user/settings.js
```diff
@@ -301,6 +301,43 @@ pimcore.settings.user.user.settings = Class.create({
...
+            fields: ['code', 'text'],
+            data : validLocales
+        });
+
+        generalItems.push({
+            xtype: 'combo',
+            fieldLabel: t('datetime_locale'),
+            typeAhead: true,
+            value: this.currentUser.datetimeLocale ?? '',
+            mode: 'local',
```
markus-moser: Let's add a `queryMode: 'local'` here as this improves the behaviour of the typeahead search. I'm not sure, but does the `mode` option even exist or should it be replaced with `queryMode`?

markus-moser:
The version preview does not respect the `datetimeLocale`. Would it be a big challenge to apply it there too?

image

public/js/pimcore/settings/user/user/settings.js
kingjia90: indeed it's not working, fixed on https://github.com/pimcore/admin-ui-classic-bundle/pull/554/commits/2534755886aaedf09c44d8378197760cbc28e1fa 
fixed also from where i have copied from, but should revisit all the occurrence/usage of this outdated `mode` in a separate PR

kingjia90: Will have a deeper look, opened an issue for it https://github.com/pimcore/admin-ui-classic-bundle/issues/569

kingjia90:
> The version preview does not respect the `datetimeLocale`. Would it be a big challenge to apply it there too?
> 
> image

leaving notes here, thank you for the input,  yes, it may be a little tricky but would be doable with some work in the backend with the twig and Carbon, maybe better in a follow up PR 
https://github.com/pimcore/admin-ui-classic-bundle/blob/9817aa7ef45734e34a7cb1cd9a6f4cf5e7f4cd4a/templates/admin/data_object/data_object/diff_versions.html.twig#L32-L34

PR #559, opened 2024-06-11
Field type "Table": add button to clone row

BlackbitDevs:
The field type `Table` has 2 buttons to "copy" and paste but this cannot be used to copy/clone a row. You can copy the table data but in the paste dialog, the dialog window immediately closes after pasting, so you cannot duplicate a row (only if you copy table content, paste it in an external editor, duplicae the desired row and then use paste button - usability nightmare ;-) ).

This PR adds a button to clone the selected row.

Alternatively / additionally we could change current behaviour of immediately closing the paste dialog box after pasting. But still cloning a row with one button click would be a better UX.

And this PR cleans up the add / delete buttons to be in the same order for columns and rows - before it was 
| Add column | Delete column | Delete row | Add row |
|---|---|---|---|

Now it is
| Add column | Delete column | Add row | Delete row |
|---|---|---|---|

### DIFF
--- a/public/css/icons.css
+++ b/public/css/icons.css
@@ -126,6 +126,16 @@
     background: url(/bundles/pimcoreadmin/img/flat-color-icons/overlay-google.svg) center center no-repeat !important;
 }
 
+.pimcore_icon_overlay_clone:before {
+    position: absolute;
+    width: 14px;
+    height: 14px;
+    bottom: 0px;
+    right: 0px;
+    content: "";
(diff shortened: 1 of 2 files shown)


mcop1:
I agree on the topic of the closing paste dialog, dont like it at all. Additionally we could ad some tooltips/alt texts for the icons. This should be a separate pr/issue though. 
But even with those changes the clone button still makes perfectly sense to me- thanks for the contribution.

PR #561, opened 2024-06-11
[Bug]: Fix Quantity Value Range turning 0 into empty/null

### LINKED ISSUE #16401: [Bug]: Data Component Quantity Value Range does not allow 0
### Pimcore version

11.1.3

### Steps to reproduce

1. Create a Class with Data Component -> Quantity Value Range
2. Create an object based on this Class and set Minimum to 0 
3. Save and reload Layout.

The Minimum-Value is empty and resave throws an exception: 

`Validation failed: Invalid dimension unit data: `

This also happens if minimum is negative and maximum is 0
This behaviour does not occur with Data Components Number and Quantity Value.

### Actual Behavior

The 0 is not displayed in input fields of "Quantity Value Range" after open or reload a data object with this type of element.

### Expected Behavior

The "0" must also be displayed in "Quantity Value Range" input fields after open or reload a data object with this type of element.

kingjia90:
Resolves https://github.com/pimcore/pimcore/issues/16401

### DIFF
--- a/public/js/pimcore/object/tags/quantityValueRange.js
+++ b/public/js/pimcore/object/tags/quantityValueRange.js
@@ -78,14 +78,14 @@ pimcore.object.tags.quantityValueRange = Class.create(pimcore.object.tags.abstra
 
         if (this.data) {
             if (!isNaN(this.data.minimum)) {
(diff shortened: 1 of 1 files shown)


lukmzig:
LGTM :)

PR #566, opened 2024-06-13
Field type Table: Support autoSize columns

BlackbitDevs:
In contrast to "Structured table", the field type table does not support configuring the "label width" (which for structured table gets used for the column widths. As columns do not have any width, they are displayed very small:
Bildschirmfoto 2024-06-13 um 10 03 59

This PR uses ExtJS's `autoSize` feature for the columns, so that the width gets automatically adjusted to their content (or column headings):
Bildschirmfoto 2024-06-13 um 10 10 29


robertSt7:
@BlackbitDevs Thanks a lot for the improvement

PR #567, opened 2024-06-13
[Improvement] Split icon library

### LINKED ISSUE #257: [Improvement] Split icon library into submenus
Latest twemoji https://github.com/pimcore/pimcore/pull/12813 introduce tons of variants based on gender, sexual orientation, skin tone etc.

Loading the icon library takes too long for JS functionality to kick in.

E.g. the feature to click on an icon to copy the path only works if all icons are loaded.

Suggestion is to split the icon library up into sub menus to make loading more efficient. 
E.g. into flat-white, flat-color, twemoji and flags.

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/257

### Additional info
image

- Split in different tabs
- Flags decreased the flags from 800 to 248
   - it was considering the language and all the variants, so multiple country flag were appearing multiple times because of their language variants
   - removed the unknown from the list as the icon would be blank
   - displaying the country name, instead of the language, sort by iso code
- Replaced `.click()` with `on('click')`  [see](https://elijahmanor.com/blog/differences-between-jquery-bind-vs-live-vs-delegate-vs-on)
- added lazy loading html attribute
- load by background image seems helping DOM to avoid refreshing (due to new image size recalculations)

##### Data Object class 
image
- moved in 2 rows (icon tools)
- added a combobox for selecting/filtering the icon type/style
- splitted the twemoji in 3 parts, as `(1/3)`
- moved the icon to be shown on the right with some indent, so that overflowing long text(icon url) doesn't overlap with the icon on the left
image

### DIFF
--- a/templates/admin/misc/icon_list.html.twig
+++ b/templates/admin/misc/icon_list.html.twig
@@ -47,10 +47,8 @@
 
         .info {
             text-align: center;
-            margin-bottom: 30px;
             clear: both;
             font-size: 22px;
-            padding-top: 50px;
         }
 
         .info small {
--- a/templates/admin/misc/icon_list.html.twig
+++ b/templates/admin/misc/icon_list.html.twig
@@ -13,8 +13,9 @@
         }
 
         .icons {
-            width:1200px;
             margin: 0 auto;
+            display:inline-block;
+            padding-top: 20px;
         }
 
         .icon {
@@ -25,9 +26,11 @@
             float: left;
             font-size: 10px;
             word-wrap: break-word;
-            cursor: copy;
             padding-top: 5px;
             box-sizing: border-box;
+        }
+        .icon img:hover{
+            cursor: copy;
         }
 
         .icon.black {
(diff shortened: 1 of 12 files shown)


templates/admin/misc/icon_list.html.twig
```diff
@@ -66,126 +71,136 @@
...
+    crossorigin="anonymous"></script>
 
 <script {{ pimcore_csp.getNonceHtmlAttribute()|raw }}>
     $( document ).ready(function() {
+
         // Add click event to copy icon path on all icons
-        $('img').each(function () {
-            $(this).click(function () {
-                pimcore.helpers.copyStringToClipboard($(this).data('imgpath'));
+        $('img, div.lazy-load').each(function () {
```
markus-moser: Just because I read about the `.on('click')` topic. I guess the whole loop could be even replaced with something like this, right? The advantage would be that you have only one event listener instead of one for each icon which could be massively more lightweight. ;)

```js
$('body').on('click', 'img, div.lazy-load', function () {
    copyStringToClipboard($(this).data('imgpath') ?? $(this).data('lazyimgpath'));
});
```

PR #568, opened 2024-06-14
Consider & showing Platform Version when installed

fashxp:
shows platform version in pimcore logo tooltip and about window if installed

depending on https://github.com/pimcore/pimcore/pull/17179 and pimcore/pimcore:11.3.0


brusch (approved):

brusch:
Tests failing because of requiring not yet released version of `pimcore/pimcore` - so should be all fine.

PR #570, opened 2024-06-18
Bugfix upload asset folder with more than 100 files

mschweiger90:
When you upload an asset folder with more than 100 files inside directly in pimcore, only 100 files are saved!


mschweiger90:

PR #572, opened 2024-06-19
Bugfix to read more than 100 files inside an uploaded folder

mschweiger90:
When you upload an asset folder with more than 100 files inside directly in pimcore, only 100 files are saved!

### DIFF
--- a/public/js/pimcore/asset/tree.js
+++ b/public/js/pimcore/asset/tree.js
@@ -363,15 +363,15 @@
                      }.bind(this));
                  } else if (item.isDirectory) {
                      // Get folder contents
-                     var dirReader = item.createReader();
-                     var entries = [];
-                     var readEntries = function() {
+                     const dirReader = item.createReader();
+                     let entries = [];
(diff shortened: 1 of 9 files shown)


robertSt7:
Hi @mschweiger90 since this is a bugfix could you please change the base branch to 1.4 and use `const` and `let` instead of `var`? Thanks

mattamon:
@robertSt7 I was able to change the base branch to 1.5 and changed to const and let.

You can ignore the sonarcloud warnings and the files changed because of the CS fixer

kingjia90:
Looks like some Chrome limitation Screenshot from 2024-09-13 13-14-40

Is it limited to Chrome v77? Could you please share your current browser version where you encountered this issue?

So the PR looks LGTM on paper, i will give it a try to see if they changed it later on

robertSt7:
@mschweiger90 Since this is only for an old Chrome version, we can't merge this. If you are able to reproduce this on newer versions, please feel free to open the PR again. Thanks for your understanding.

PR #580, opened 2024-06-24
composer.json: use pimcore/pimcore: ^11.3.0-RC1 to still allow 11.x-dev

brusch:
When having a fixed contraint to `11.3.0-RC1` it blocks all "newer" branches and tags to install, such as `11.x`-dev.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
     "cbschuld/browser.php": "^1.9.6",
     "phpoffice/phpspreadsheet": "^1.24 || ^2.1",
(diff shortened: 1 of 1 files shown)


fashxp:
we should move to `^11.3.0` anyway, right? 
our tests should install `11.3.0-RC1` in case there is no `11.3.0` yet, right?

brusch:
@fashxp thought so too, but it seems it doesn't like it 😉 
```
    - Root composer.json requires pimcore/pimcore ^11.3.0, found pimcore/pimcore[v11.3.0-RC1, 11.x-dev] but it does not match your minimum-stability.
```

So better leave it as it was for now, doesn't hurt.

PR #581, opened 2024-06-24
Fix: Usage of explode() on an empty string

blankse:
See https://github.com/pimcore/pimcore/pull/17221

Line 760 creates an array `['']` but should be `[]` when the additional domains textarea was empty.

### DIFF
--- a/src/Controller/Admin/Document/DocumentController.php
+++ b/src/Controller/Admin/Document/DocumentController.php
@@ -757,7 +757,7 @@ class DocumentController extends ElementControllerBase implements KernelControll
     {
         $domains = $request->get('domains');
         $domains = str_replace(' ', '', $domains);
-        $domains = explode("\n", $domains);
+        $domains = $domains ? explode("\n", $domains) : [];
 
         if (!$site = Site::getByRootId((int)$request->get('id'))) {
             $site = Site::create([
--- a/src/Controller/Admin/Document/DocumentController.php
+++ b/src/Controller/Admin/Document/DocumentController.php
@@ -755,7 +755,7 @@ class DocumentController extends ElementControllerBase implements KernelControll
      */
     public function updateSiteAction(Request $request): JsonResponse
     {
-        $domains = $request->request->get('domains', '');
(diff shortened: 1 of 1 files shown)


src/Controller/Admin/Document/DocumentController.php
```diff
@@ -755,13 +755,13 @@ public function publishVersionAction(Request $request): JsonResponse
      */
     public function updateSiteAction(Request $request): JsonResponse
     {
-        $domains = $request->get('domains');
+        $domains = $request->request->get('domains', '');
         $domains = str_replace(' ', '', $domains);
-        $domains = explode("\n", $domains);
+        $domains = $domains ? explode("\n", $domains) : [];
```
kingjia90, line 760: what about 
```
        $domains = $request->request->get('domains');
        if ($domains){
           if (is_string($domains)){
               $domains = str_replace(' ', '', $domains);
               $domains = $domains ? explode("\n", $domains) : [];
           }
        }else{
           $domains = [];
        }
```
?

blankse: @kingjia90 I changed it to getString(), so it must be a string. See https://github.com/pimcore/admin-ui-classic-bundle/pull/581/commits/f616ff80df6407638922ca17e0c8b89d7ab36e83

blankse:
@mattamon Can you review this too? Otherwise https://github.com/pimcore/pimcore/pull/17221 will case an error with empty additional domains textarea.

The milestone in https://github.com/pimcore/pimcore/pull/17221 is also missing. Can you add it?
Why did you lock all PRs after merging?

mattamon:
> @mattamon Can you review this too? Otherwise [pimcore/pimcore#17221](https://github.com/pimcore/pimcore/pull/17221) will case an error with empty additional domains textarea.
> 
> The milestone in [pimcore/pimcore#17221](https://github.com/pimcore/pimcore/pull/17221) is also missing. Can you add it? Why did you lock all PRs after merging?

Locking is done automatically, sorry if it causes inconvenience.

Issue #603, opened 2024-07-04
[Bug]: When the classification store column is filtered on grid objects that inherit value from parent are not returned


pfilipkowskilemon:
Steps to reproduce
1. Open grid and add column from classification store
2. Try to filter classification store column

Actual Behavior
When the classification store column is filtered on grid only objects that have this value assigned are returned. But what about objects that have this value inherited from parent?

Expected Behavior
When the classification store column is filtered on grid objects that have this value assigned and objects that inherited this value from parent are returned

fashxp:
Hi there, as we begin sunsetting admin-ui-classic in favor of Pimcore Studio, we won’t be able to fix or implement this issue. We'll keep an eye on it during Studio development though. 
Closing it out to keep things tidy – really appreciate your contribution and understanding!

PR #604, opened 2024-07-04
bugfix - show filtered inherited cs values on grid

pfilipkowskilemon:
https://github.com/pimcore/admin-ui-classic-bundle/issues/603


pfilipkowskilemon:

robertSt7:
Hi @pfilipkowskilemon Since this is a bugfix, could you please change the base branch to 1.5? Thanks

robertSt7:
Hi @pfilipkowskilemon thanks for changing the branch. 
This approach only works across two levels but not across three levels of inheritance. Could you please also take a look at this? Thx
image

pfilipkowskilemon:
Hi @robertSt7  I apologize for the late reply. You are correct about the fact that it only works for two levels. This solution fits for my client so I thought it might be useful for others.
Unfortunately, I have no idea how to improve this mechanism. So if you want to close PR then there is no problem.

robertSt7:
@pfilipkowskilemon no worries. Sorry we don't have time at the moment to have a deeper look at this. We will come back to this later. Maybe you find a solution or some of the other community member has an idea in the meanwhile, we would highly welcome this.

robertSt7:
@pfilipkowskilemon I think you override your commits, could you please restore them? Thanks

pfilipkowskilemon:
@robertSt7 I don't know exactly what happened, probably some github things, because I didn't do anything with the fork myself in december. The original commits from 04.07.2024 are now visible but I think there should be a target branch change to 2.x? I did sync my fork and pushed the original commits into it. If something is wrong with it, please let me know.

robertSt7:
@pfilipkowskilemon  We discussed this internal and decided to not merge this, because it leads to some weird behaviour in the filtering with 3 levels. We will fix this in studio-ui. Thanks for your understanding

PR #606, opened 2024-07-08
Run some workflows only in pimcore repos

herbertroth:


blankse:
@herbertroth Do this work with PRs from forks?

herbertroth:
@blankse This disables CS Fixer runs on PRs from forks. It’s simply not possible to do this on PRs for forks. This would require a way to get write permissions on the forked repo with a token stored at Pimcore. For very good reasons, this can’t work. So the CS Fixer will only run for PRs for a Pimcore-owned repo to avoid error messages indicating that a token is not present.

blankse:
@herbertroth But with the `pull_request_target` it was working with PRs from forks.

For example see here: https://github.com/pimcore/pimcore/pull/17012
This PR is from my fork and the php-cs-fixer did make changes.

herbertroth:
@blankse This PR solves a different problem. If you fork this to E.G.: blankse/pimcore and than do a PR to blankse/pimcore this will run and try to find a key stored at pimcore. The key will not be present and will trigger an error. If you do a PR from blankse/pimcore to pimcore/pimcore this will work.

blankse:
@herbertroth Yes, with `pull_request_target` it should work :)

herbertroth:
@blankse I would be happy if we could test this together after the merge. It’s tricky to do this alone.

PR #607, opened 2024-07-08
Fixed an issue of workflow permissions conflict with keybindings

rahul-pim-dev:
Fixed an issue of keybinding conflict between workflow permissions

Reported issue : https://github.com/pimcore/pimcore/issues/17095


markus-moser:
Thank you for your PR. Unfortunately we still would have problems if we solve it like this (e.g. that it only works for data objects but not for assets and documents).

Therefore I created a different PR with a more centralized approach: 
https://github.com/pimcore/pimcore/pull/17373

PR #614, opened 2024-07-15
Fix nice path support inheritance

### LINKED ISSUE #591: [Bug] Route "pimcore_admin_element_getnicepath" does not support inheritance
The route "pimcore_admin_element_getnicepath" does not support inheritance. The columns are left empty:

image
image

After I adjust the code in file `src/Controller/Admin/ElementController.php` as follows:
```diff
        if ($request->request->getBoolean('loadEditModeData')) {
            $idProperty = $request->get('idProperty', 'id');
            $methodName = 'get' . ucfirst($fieldname);
            if ($ownerType == 'object' && method_exists($source, $methodName)) {
+                DataObject\Concrete::setGetInheritedValues(true);
                $data = $source->$methodName();
+                DataObject\Concrete::setGetInheritedValues(false);

                $editModeData = $fd->getDataForEditmode($data, $source);
                // Inherited values show as an empty array
                if (is_array($editModeData) && !empty($editModeData)) {
```

Or alternatively by adding `DataObject\Concrete::setGetInheritedValues(true);` to beginn of the method.

Then this works:
image
image

---

However, I am a little confused about the expected pimcore behaviour and how I can best resolve this via PR.
This commit with the comment leaves me a bit puzzled: 
https://github.com/pimcore/admin-ui-classic-bundle/commit/1acb2dd9893bc3cca14e959dcf1aa8f99ac4ca09

What is mean with the comment `Inherited values show as an empty array`?
Should inherited data for the table be processed somewhere other than here?

noahcommandeur-twocream:
Resolves #591

### DIFF
--- a/public/js/pimcore/object/tags/fieldcollections.js
+++ b/public/js/pimcore/object/tags/fieldcollections.js
@@ -64,12 +64,9 @@ pimcore.object.tags.fieldcollections = Class.create(pimcore.object.tags.abstract
             allowedTypes: allowedTypes.join(","),
             object_id: this.object.id,
             field_name: this.fieldConfig.name,
+            layoutId: this.object.data.currentLayoutId,
             forObjectEditor: 1
         };
-
-        if (typeof this.fieldConfig.layoutId !== "undefined") {
-            extraParams.layoutId = this.fieldConfig.layoutId;
-        }
 
         Ext.Ajax.request({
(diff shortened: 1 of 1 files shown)


fashxp:
thx very much for the PR. 
since this is a bugfix, please rebase on 1.5.

PR #615, opened 2024-07-15
[Improvement]: Add download button for a asset relation

kevinloeber-twocream:

A small feature to download assets faster. A download button is added to a ManyToOneRelation. This button appears, but only if the setting `allow inline download` is activated in the class definition. 

## Additional info

With this function you have the possibility to download the assets faster instead of opening or searching the asset first and then clicking on download.

image
image

### DIFF
--- a/public/js/pimcore/object/tags/manyToOneRelation.js
+++ b/public/js/pimcore/object/tags/manyToOneRelation.js
@@ -286,7 +286,7 @@ pimcore.object.tags.manyToOneRelation = Class.create(pimcore.object.tags.abstrac
 
         this.composite = Ext.create('Ext.form.FieldContainer', compositeCfg);
 
-        if (this.fieldConfig.assetsAllowed && this.name.endsWith('DataSheet')) {
+        if (this.fieldConfig.assetInlineDownloadAllowed) {
             this.composite.remove(
                 this.composite.items.getAt(4)
             );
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/manyToOneRelation.js
```diff
@@ -286,6 +286,22 @@ pimcore.object.tags.manyToOneRelation = Class.create(pimcore.object.tags.abstrac
 
         this.composite = Ext.create('Ext.form.FieldContainer', compositeCfg);
 
+        if (this.fieldConfig.assetsAllowed && this.name.endsWith('DataSheet')) {
```
mcop1: Maybe I am missing something, but `this.name` reflects the name of the field. Why is there a check for `DataSheet`?

kevinloeber-twocream: I had probably still implemented it for pimcore 10, changed the query

mcop1: Thanks, but isnt it the same as https://github.com/pimcore/admin-ui-classic-bundle/blob/d53a2cb0ba8c64b6833467f13fbe72e90c3f97a7/public/js/pimcore/object/tags/manyToOneRelation.js#L223 then?

kevinloeber-twocream: Yes, that is correct. I suspect that the function is already included in Pimcore. I will discuss it again internally.

mcop1:
Thanks @kevinloeber-twocream, I will close the pr for now, feel free to reopen as soon as you have new feedback.

PR #618, opened 2024-07-17
fix: search for e-mails does not break with a leading at-sign

fabfischer:
the (email) search throws an error when searching with a leading at-sign (@)
see pimcore-demo:
CleanShot 2024-07-17 at 09 21 49@2x

this fixes it (removes the @ character, which should return results with the given domain).


fabfischer:

markus-moser:
Thanks a lot 👍

markus-moser:
As it is a bugfix, I'll add it to the 1.5 branch too.

PR #619, opened 2024-07-17
Support frame-anchestors property in CSP headers.

elements-team:
# Pull Request

## Problem
iFrames can load insecure content from untrusted URLs.

## Solution
By adding the frame-anchestor property only trusted URLs can be loaded in iFrames. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors This improves the security and minimizes external impacts.


robertSt7:
Hi @elements-team we think this is a bug-fix. Could you please re-base the PR to 1.5 and add some upgrade notes? TIA

PR #622, opened 2024-07-19
Bug document Save Buttons

### LINKED ISSUE #592: Bug document Save Buttons
When a document gets published it is not possible anymore to "save only" without publishing. this behaviour can be tested in the pimcore demo

image

robertSt7:
resolves #592

### DIFF
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -246,6 +246,13 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                 }
 
                 if (this.toolbarButtons.save && this.toolbarButtons.publish) {
+                    if (this.isAllowed("save")) {
+                        this.toolbarButtons.publish.menu.items.items[1].setHidden(false)
+                    }
+                    if (this.isAllowed("settings")) {
+                        this.toolbarButtons.publish.menu.items.items[2].setHidden(false)
+                    }
+
                     this.toolbarButtons.publish.show();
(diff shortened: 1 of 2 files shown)


public/js/pimcore/document/document.js
```diff
@@ -246,6 +246,19 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                 }
 
                 if (this.toolbarButtons.save && this.toolbarButtons.publish) {
+                    if (this.isAllowed("save")) {
+                        const menuItem = this.toolbarButtons.publish.menu.items.items.find(
+                            element => element.text === t('save_draft')
+                        )
+                        menuItem.setHidden(false)
```
kingjia90, line 253: looks like this condition `!this.data.published` could be tweaked instead 🤔 

https://github.com/pimcore/admin-ui-classic-bundle/blob/cf412952e205a9b0a6000bf61fdfdea47b1b8bab/public/js/pimcore/document/page_snippet.js#L265

but must be careful of https://github.com/pimcore/pimcore/issues/5136#issuecomment-547485870

is it affected by https://github.com/pimcore/admin-ui-classic-bundle/blob/12309d931e100b2c8f91a0986759e94e12f8a269/public/js/pimcore/document/document.js#L237 ?

kingjia90: ahh, nevermind, it's only happening "on the fly" after publishing

Issue #627, opened 2024-07-22
treeGetChildrenByIdAction has n+1 select issues


cancan101:
Accessing `treeGetChildrenByIdAction`  (`/tree-get-children-by-id`) exhibits n+1 select behavior:
Using data from Sentry:
![image](https://github.com/user-attachments/assets/f4411165-5d88-4200-a0ea-31b23290bbbb)
The query handed to https://github.com/pimcore/admin-ui-classic-bundle/blob/7999ef435d84f133d469539b5664c6b1d65c176e/src/Controller/Admin/TagsController.php#L155-L182

looks to load lazily. something like may address the issue: https://www.doctrine-project.org/projects/doctrine-orm/en/3.2/reference/working-with-objects.html#by-eager-loading

kingjia90:
Thank you for reporting!
I've opened 2 PRs that may help decreasing these queries, please see https://github.com/pimcore/pimcore/pull/17390 and https://github.com/pimcore/admin-ui-classic-bundle/pull/629

Could you please apply the changes and re-run Sentry to see if there's an improvement? Thank you in advance

kingjia90:
Marking this as done as the 2 PRs (https://github.com/pimcore/pimcore/pull/17390 and https://github.com/pimcore/admin-ui-classic-bundle/pull/629) that might improve the behavior are merged

Let me know if you find anything else that we may improve or if it's not resolved and still occurs.

cancan101:
I am still seeing N+1 issues on admin bundle 1.7.3 and Pimcore v11.5.2:

<img width="1132" alt="Image" src="https://github.com/user-attachments/assets/cfee803c-c97f-4ba5-b7c5-aa482ad88952" />

kingjia90:
Thank you for the follow up!

I have took a second look and didn't find anything that can be done further yet, just wondering, as i am not familiar with Sentry, is it by any chance that the runtime cache is disabled when running this? 

Does the test result have any difference by reverting the changes in the PR that i have provided in the past? 

Thank you in advance

cancan101:
Sentry should not affect caching behavior.

For us the majority of the tags are nested under a top level. Not sure if that will affect how the caching was built:

![Image](https://github.com/user-attachments/assets/21fd89aa-ecd5-4a1c-be60-60c043177d3b)
(there are ~2000 tags under "AzureCV")

cancan101:
[...]

PR #629, opened 2024-07-24
[Task]: Optmize tags treeGetChildrenByIdAction

kingjia90:
Related https://github.com/pimcore/admin-ui-classic-bundle/issues/627

Slightly reduce (at least by 2) how many times will be querying the entire tags listing multiple times via `hasChildren` and `getChildren` when no children are there, see also
https://github.com/pimcore/pimcore/blob/e74e2b4be10bb7abf75f0c95f3591e509f40b273/models/Element/Tag.php#L293-L297
https://github.com/pimcore/pimcore/blob/e74e2b4be10bb7abf75f0c95f3591e509f40b273/models/Element/Tag.php#L320-L323


lukmzig (approved):
LGTM

PR #635, opened 2024-07-25
[PIMCORE-17387] Fix bug with reorder of localized fields in class

nbogucki:
[#17387](https://github.com/pimcore/pimcore/issues/17387)

Due to last changes in the file `public/js/pimcore/helpers.js` at line 3369 was added method is ComponentAsChildAllowed. This method return false for localizedfields because fields types were not included in `public/js/pimcore/object/helplers/layout.js` in getRawAllowedTypes method. So I overwrited localizedfields key with 'data' type. After this changes localizedfields works correctly.


nbogucki:

markus-moser:
Thanks a lot! 
The same problem occured for the block data type too, therefore I added the `data` type also to the block.

PR #636, opened 2024-07-30
Fix ManyToOne relation width with combo display mode

blankse:
Followup to https://github.com/pimcore/admin-ui-classic-bundle/commit/946778db2ecf0334d3696cbddd7ef652b8fc2c4a

The combo box only opens via the arrow, which is difficult to click on as it is half hidden.

Before:
image

After:
image


robertSt7:
@blankse Thanks for the fix

PR #639, opened 2024-07-30
Add force parameter to retrieve object directly from database

croitoruion:
The object from the database should be retrieved using the force parameter. Currently, it can be fetched from the cache, which may result in outdated (cached) data.
We encountered a bug where the system returns a draft version of objects instead of the latest saved version. This happens because $objectFromDatabase is compared to a cached previous version, not the actual version from the database. 
Adding the force parameter ensures the object is fetched directly from the database.
image

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectController.php
+++ b/src/Controller/Admin/DataObject/DataObjectController.php
@@ -293,7 +293,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     public function getAction(Request $request, EventDispatcherInterface $eventDispatcher, PreviewGeneratorInterface $defaultPreviewGenerator): JsonResponse
     {
         $objectId = $request->query->getInt('id');
-        $objectFromDatabase = DataObject\Concrete::getById($objectId);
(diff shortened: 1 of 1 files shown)


croitoruion:

mhgtlc:
@brusch @wisconaut  can you please merge this PR?

fashxp:
Could you please outline more in detail what happened on your end? 
Fetching data from DB will affect performance and at least in theory, the cached data should always be up-to-date. So probably there is another reason for that problem?

fashxp:
:ping_pong: @mhgtlc @croitoruion

fashxp:
@mhgtlc @croitoruion any updates?

croitoruion:
> @mhgtlc @croitoruion any updates?

We updated the Redis clusters and cleared all Pimcore caches, and the problem disappeared. I believe we encountered a very edge case. We're considering that when retrieving an object from the database, it should be freshly fetched and that fixed the problem we faced, understanding that this may affect performance.
We can close this PR for now. 
Thanks

lukmzig:
@croitoruion thank you very much for the clarification. Closing the PR based on your comment :+1:

Issue #642, opened 2024-07-31
[Bug] Data object color field type doesn't respect label align


blankse:
# Steps to reproduce

1. Create a class "Test"
2. Create a Panel with label align "top"
![image](https://github.com/user-attachments/assets/909cbab7-d31f-47c9-ab11-e3cbc702a728)
3. Create a Color field in the panel
![image](https://github.com/user-attachments/assets/8cc193ef-d23e-4479-abcc-47b25f57343a)
4. Create a object
=> Label is "left" and not "top"
![image](https://github.com/user-attachments/assets/f6969d94-fccf-49b6-a383-f4422cca7a60)

blankse:
Fixed by https://github.com/pimcore/admin-ui-classic-bundle/pull/643

PR #643, opened 2024-07-31
Fix: Data object color field type doesn't respect label align

blankse:
Fix: https://github.com/pimcore/admin-ui-classic-bundle/issues/642

After:
image


mcop1:
Thank you!

PR #647, opened 2024-08-05
Fix: Missing objectId context in gridCellEditor

blankse:
1. Create a manyToOne field with combo view mode
2. Create a AdminSearchEvents::SEARCH_LIST_BEFORE_FILTER_PREPARE event to add filters to the drop down options
3. In the editmode the context have the objectId but in the gridCellEditor this context is missing

example event:
```
<?php

declare(strict_types=1);

namespace App\EventListener;

use Pimcore\Bundle\SimpleBackendSearchBundle\Event\AdminSearchEvents;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\EventDispatcher\GenericEvent;

class AdminSearchListener implements EventSubscriberInterface
{
    public static function getSubscribedEvents(): array
    {
        return [
            AdminSearchEvents::SEARCH_LIST_BEFORE_FILTER_PREPARE => 'onBeforeFilterPrepare',
        ];
    }

    public function onBeforeFilterPrepare(GenericEvent $event): void
    {
        $requestParams = $event->getArgument('requestParams');
        $context = json_decode($requestParams['context'], true);
    }
}

```

### DIFF
--- a/public/js/pimcore/element/helpers/gridCellEditor.js
+++ b/public/js/pimcore/element/helpers/gridCellEditor.js
@@ -36,40 +36,39 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
 
         value = Ext.clone(value);
 
-        var fieldInfo = Ext.clone(this.config.fieldInfo);
-        var fieldType = this.config.elementType;
+        const fieldInfo = Ext.clone(this.config.fieldInfo);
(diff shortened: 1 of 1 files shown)


lukmzig:
Thank you very much for the fix @blankse

PR #648, opened 2024-08-06
Add missing label defaults to numeric control

cancan101:
Without this 0 and missing values lead to invalid CSS. These changes are consistent with input.js
e.g. you get a css width of `calc(80px + undefined)` if the labelWidth in parent is 0 or not specified.


robertSt7:
@cancan101 Thanks for making it consistent

PR #653, opened 2024-08-07
[Bug]: Fix print document save error

### LINKED ISSUE #652: [Bug]: Print document save error
### Pimcore version

Platform 2024.2

### Steps to reproduce

- open any print-document https://demo.pimcore.fun/admin/login/deeplink?document_79_printpage
- save and publish
- save error popup 
Bildschirmfoto 2024-08-07 um 11 15 51

Reason:
Print-document has not Submenu-Button 'Save only scheduled tasks', but js tries to hide this button after successfull saving.

### Actual Behavior

save error

### Expected Behavior

save without error

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/652
Regression https://github.com/pimcore/admin-ui-classic-bundle/pull/622

A [console.log(e)](https://github.com/pimcore/admin-ui-classic-bundle/blob/7cad2aa699dc350fcdb1d31161adc3661dab91af/public/js/pimcore/document/document.js#L174) here revelead that not every document type has these buttons and could be caught by `try` block and display a false negative save notifcation. The save should be considered successful, so no data loss.

image

### DIFF
--- a/public/js/pimcore/document/document.js
+++ b/public/js/pimcore/document/document.js
@@ -250,13 +250,13 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                         const menuItem = this.toolbarButtons.publish.menu.items.items.find(
                             element => element.text === t('save_draft')
                         )
-                        menuItem.setHidden(false)
+                        menuItem?.setHidden(false)
                     }
                     if (this.isAllowed("settings")) {
                         const menuItem = this.toolbarButtons.publish.menu.items.items.find(
                             element => element.text === t('save_only_scheduled_tasks')
                         )
-                        menuItem.setHidden(false)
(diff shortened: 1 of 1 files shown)


kingjia90:
Mhhh, another potential bug is that the buttons are overcrowded and repeated on newly created printpage and containers
image

Could you please also check if is recent change? TIA

robertSt7:
> Mhhh, another potential bug is that the buttons are overcrowded and repeated on newly created printpage and containers image
> 
> Could you please also check if is recent change? TIA

@kingjia90 This also happens in 1.5. Should we change it in this PR or create a new issue?

kingjia90:
I'd say maybe in a new issue, it's only a visual issue and seems not recent anyways. Thank you

PR #654, opened 2024-08-07
Bugfix - Sharing grids to users

torqdev:
Bugfix for sharing grid configs to other users and setting as favourite.

Can be replicated on `demo.pimcore.fun`. Steps to recreate:
1. Create grid config, and navigate to 'Save & Share'.
2. Add users. Check set as favourite. Do not set 'Share globally'.
3. Save & Share

### DIFF
--- a/src/Controller/Admin/Asset/AssetHelperController.php
+++ b/src/Controller/Admin/Asset/AssetHelperController.php
@@ -604,7 +604,7 @@ class AssetHelperController extends AdminAbstractController
             $sharedUserIds = $metadata['sharedUserIds'];
 
             if ($sharedUserIds) {
-                $sharedUsers = explode(',', $sharedUserIds);
+                $sharedUsers = array_map('intval', explode(',', $sharedUserIds));
             }
         }
 
@@ -620,7 +620,7 @@ class AssetHelperController extends AdminAbstractController
         foreach ($sharedUsers as $id) {
             // Check if the user has already a favourite
             $favourite = GridConfigFavourite::getByOwnerAndClassAndObjectId(
-                (int) $id,
+                $id,
                 $gridConfig->getClassId(),
                 0,
                 $gridConfig->getSearchType()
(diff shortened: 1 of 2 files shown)


martineiber (changes requested):
Thank you for this fix. LGTM :+1: 

I have one more Request. Could you also change this for the AssetGrid here?
https://github.com/pimcore/admin-ui-classic-bundle/blob/7554a0fc3fd89d69e6c90fadf62d3f3edaa03d56/src/Controller/Admin/Asset/AssetHelperController.php#L649

blankse:
I think we could change the line 978 to:
` $sharedUsers = array_map('intval', explode(',', $sharedUserIds));`

So we can remove lot of (int) casts.

Right?

martineiber:
> I think we could change the line 978 to: ` $sharedUsers = array_map('intval', explode(',', $sharedUserIds));`
> 
> So we can remove lot of (int) casts.
> 
> Right?

This would also be a nice solution :)

src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -636,7 +636,7 @@ protected function updateGridConfigFavourites(?GridConfig $gridConfig, array $me
                     }
 
                     // Check if the user is the owner. If that is the case we do not update the favourite
-                    if ((int) $favouriteGridConfig->getOwnerId() === (int) $id) {
+                    if ((int) $favouriteGridConfig->getOwnerId() === $id) {
```
blankse: getOwnerId() can only be int or null. So there is also no convert needed.
```suggestion
                    if ($favouriteGridConfig->getOwnerId() === $id) {
```

src/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -1008,15 +1008,15 @@ protected function updateGridConfigFavourites(?GridConfig $gridConfig, array $me
                     }
 
                     // Check if the user is the owner. If that is the case we do not update the favourite
-                    if ((int) $favouriteGridConfig->getOwnerId() === (int) $id) {
+                    if ((int) $favouriteGridConfig->getOwnerId() === $id) {
```
blankse:
```suggestion
                    if ($favouriteGridConfig->getOwnerId() === $id) {
```

martineiber (approved):
LGTM :+1: 
Thank you for your contribution

PR #656, opened 2024-08-13
Show tree search if paging is not necessary

BlackbitDevs:
If you set 
```yaml
pimcore:
  objects:
    tree_paging_limit: 1000
```

and an element has less than 1000 children, the paging toolbar does not get displayed. The downside is that also the search does not get displayed then.

This PR changes this: Even if all items fit on the current page, there is still the search shown:
Bildschirmfoto 2024-08-13 um 08 13 55

The behaviour if `tree_paging_limit` < total number of children does not get changed.


fashxp:
so search is now shown all the time for every folder?

BlackbitDevs:
Yes. Of course we can also add a threshold so that the search only gets shown if there are more than X items.

mattamon:
> Yes. Of course we can also add a threshold so that the search only gets shown if there are more than X items.

I think a threshold would be nice yes. Could use the default limit again like 30?

BlackbitDevs:
Have added threshold: only if parent element has >= 30 child items, the search does get shown.

mattamon:
> Have added threshold: only if parent element has >= 30 child items, the search does get shown.

Perfect thank you!
Since this is going to be in the 1.6 version could you also add the changes to the CHANGELOG.md so it is documented? 
That would be awesome!

BlackbitDevs:
Added changelog.

mattamon (approved):
Looks good to me! Thank you for your effort! 

@fashxp WDYT? The 30 element search limit was added.

public/js/pimcore/overrides.js
```diff
@@ -400,15 +400,11 @@ Ext.define('pimcore.data.PagingTreeStore', {
...
 
-            if (me.pageSize < total || node.inSearch) {
-                node.needsPaging = true;
-                node.pagingData = {
-                    total: data.total,
-                    offset: data.offset,
-                    limit: data.limit
-                }
-            } else {
-                node.needsPaging = false;
```
kingjia90, line 411: Casually coming from https://github.com/pimcore/admin-ui-classic-bundle/issues/840, is it that `needsPaging` is always true (everywhere) after this PR?

@BlackbitDevs do you recall why it got changed like this? I don't see any place where it could be 'false', or am i missing anything? thank you in advance ☺️

BlackbitDevs: @kingjia90 Am not sure. The goal was to show the tree search even if the number of child items does not exceed the paging limit. If this is still the case with your changes from #850, everybody is happy ;-)

PR #659, opened 2024-08-19
Relational fields: Add column filtering

BlackbitDevs:
There already is a *search* feature to filter shown items of relational fields:
Bildschirmfoto 2024-08-19 um 11 11 18

This PR extends filtering relational fields by adding support for column filters:
Bildschirmfoto 2024-08-19 um 11 12 49


public/js/pimcore/object/tags/advancedManyToManyRelation.js
```diff
@@ -117,6 +117,7 @@ pimcore.object.tags.advancedManyToManyRelation = Class.create(pimcore.object.tag
 
         var visibleFieldsCount = columns.length;
 
+        var filterType = 'list';
```
mcop1:
```suggestion
        let filterType = 'list';
```

mcop1 (commented):
Other than the pr looks fine to me, thanks!

public/js/pimcore/object/tags/advancedManyToManyRelation.js
mcop1: Would then be consistent with https://github.com/pimcore/admin-ui-classic-bundle/pull/659/files#diff-bea40e721430c1920c766df73237226f86c16a40e65012c1051c2cdff1ee8eaeR160 too

mcop1:
Thanks again :+1:

PR #662, opened 2024-08-19
Publish for documents "failed"

storbahn:
If menuItem is null setHidden will fail and only the error message "saving_failed" will be shown.


blankse:
@storbahn Was fixed already by https://github.com/pimcore/admin-ui-classic-bundle/pull/653, right?

lukmzig:
@storbahn thank you very much for your contribution!
We are closing the PR as it seems to be duplicate of https://github.com/pimcore/admin-ui-classic-bundle/pull/653. If this is still not resolved, please feel free to re-open.

@blankse thank you for pointing out the other PR.

PR #663, opened 2024-08-19
Added grid preview for field collections

fashxp:
alternative for https://github.com/pimcore/admin-ui-classic-bundle/pull/388

details see https://github.com/pimcore/pimcore/pull/17510

### DIFF
--- a/public/js/pimcore/object/tags/fieldcollections.js
+++ b/public/js/pimcore/object/tags/fieldcollections.js
@@ -53,32 +53,53 @@ pimcore.object.tags.fieldcollections = Class.create(pimcore.object.tags.abstract
                     }
 
                     let preview = '';
+                    let plainText = false;
+                    let fieldCollectionItems = record.data[key];
 
-                    const fieldCollectionItems = record.data[key];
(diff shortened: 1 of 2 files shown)


public/js/pimcore/object/tags/fieldcollections.js
```diff
@@ -48,8 +48,34 @@ pimcore.object.tags.fieldcollections = Class.create(pimcore.object.tags.abstract
...
+                    if(typeof record.data[key] === 'string') {
+                        return record.data[key];
+                    }
+
+                    let preview = '';
+
+                    const fieldCollectionItems = record.data[key];
+
+                    let previousFieldCollectionItemType = null;
+                    for (let fieldCollectionItem of fieldCollectionItems) {
```
kingjia90, line 65: LGTM but just casually noticed that the preview is not working properly on this, as it displays `[object Object]`
image

kingjia90: with https://github.com/pimcore/admin-ui-classic-bundle/pull/663/commits/8e9fe514b9bcf3f83bb440c85dde6173c8a5aa40 will look like image

PR #664, opened 2024-08-20
[Object editing] Relational fields' grid: remember column widths

BlackbitDevs:
When there are a lot of columns to be displayed for (advanced) many-to-many (object) relation fields, the available space may not suffice for some users while it does nor others, depending on their monitor size. Currently for the configured "visible fields" the widths of the fields gets distributed evenly:
https://github.com/pimcore/admin-ui-classic-bundle/blob/0972e3b846bfe1133b359b5fc3417d1e0e40d622/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L375
And for meta fields of advanced relations you can only set the width globally for all users.

Of course, the user can adjust column widths when editing an object but after reloading the object or opening another object of the same class, he has to do this again and again.

With this PR the column widths get saved to the browser's local storage and restored when loading an object of the same class and layout.


mattamon:
Hey @BlackbitDevs could you please resolve the conflict.

BlackbitDevs:
Conflict resolved.

mattamon (approved):
Works and is a nice addition! 

Thank you for the contribution!

PR #670, opened 2024-08-28
[Bug, EC]: Saving a grid view makes read-only columns visually editable

kingjia90:
Resolves https://pimcore.atlassian.net/browse/PEES-376

Steps to reproduce (from @torqdev ):

- Create demo user, giving on the objects, quantity value units, and select options permissions. Under workspaces, the user has all permissions for the /Product Data/Cars folder.
- Create custom layout on Car DataObject. For instance this one is calledLimitedAccessCarLayout. This layout has a small set of fields, including the carClass select field that is marked not editable.
- Assign this custom layout to be the only available layout for the DemoUser created above.
- Open incognito tab, sign in as DemoUser, navigate to the /Product Data/Cars folder in the DataObjects section.
- The default grid should follow the the assigned custom layout fields and specifications. In this case they do, and we receive a warning.
- Create and save an updated grid config, reload the grid, select the updated grid config.
- Fields are now editable

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/src/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -360,12 +360,9 @@ class DataObjectHelperController extends AdminAbstractController
                             'key' => $key,
                             'type' => 'system',
                             'label' => $key,
-                            'locked' => $sc['locked'] ?? null,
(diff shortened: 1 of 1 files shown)


mattamon (approved):
LGTM!

PR #678, opened 2024-09-02
Support frame-anchestors property in CSP headers.

elements-team:
# Pull Request

## Problem
iFrames can load insecure content from untrusted URLs.

## Solution
By adding the frame-anchestor property only trusted URLs can be loaded in iFrames. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors This improves the security and minimizes external impacts.


elements-team:
@robertSt7 sorry for the circumstances. I accidentially closed https://github.com/pimcore/admin-ui-classic-bundle/pull/619.
So I reopened a new PR including the changes that you requested (Changelog + rebase onto 1.5).

BR

robertSt7:
@elements-team Thanks for finding and taking care of this security improvement

PR #681, opened 2024-09-05
Fix resolve display issue in grid view for filter by link feature

noahcommandeur-twocream:
In the Grid View, the field is sometimes cut off in relation fields in the Filter by link option; this is corrected by this.

image


mattamon (approved):
@noahcommandeur-twocream another thank you for fixing the display here!

PR #682, opened 2024-09-06
Fix date range fields read only

noahcommandeur-twocream:
You could not set DateRange Fields ReadOnly before because an ExtJS error occurred when calling them

### DIFF
--- a/public/js/pimcore/object/tags/dateRange.js
+++ b/public/js/pimcore/object/tags/dateRange.js
@@ -82,8 +82,8 @@ pimcore.object.tags.dateRange = Class.create(pimcore.object.tags.abstract, {
 
     getLayoutShow: function () {
         this.component = this.getLayoutEdit();
-        this.component.items[0].setReadOnly(true);
-        this.component.items[2].setReadOnly(true);
+        this.component.items.items[0].setReadOnly(true);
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tags/dateRange.js
```diff
@@ -82,8 +82,8 @@ pimcore.object.tags.dateRange = Class.create(pimcore.object.tags.abstract, {
 
     getLayoutShow: function () {
         this.component = this.getLayoutEdit();
-        this.component.items[0].setReadonly(true);
-        this.component.items[2].setReadonly(true);
+        this.component.items[0].setReadOnly(true);
+        this.component.items[2].setReadOnly(true);
```
robertSt7:
```suggestion
        this.component.items.items[0].setReadOnly(true);
        this.component.items.items[2].setReadOnly(true);
```

For me `this.component.items` isn't an array. We need to add another `items` to get the right items.

noahcommandeur-twocream: You are absolutely right, my mistake, I had it exactly the same with my local, only I had copied the whole thing into the editor incorrectly...

robertSt7:
@noahcommandeur-twocream Thanks for the fast response and fixing this bug

PR #684, opened 2024-09-06
Fix display reverse relation field error in grid view

noahcommandeur-twocream:
Before:
image

Fix:
image
image


mattamon (approved):
@noahcommandeur-twocream thank you for fixing the display part!

PR #687, opened 2024-09-11
Remove double width in gridColumConfig after fixing display reverse relation field error in grid view

noahcommandeur-twocream:
In the Grid View, the field is sometimes cut off in relation fields in the Filter by link option; this is corrected by this.
https://github.com/pimcore/admin-ui-classic-bundle/pull/684


mattamon (approved):
@noahcommandeur-twocream thank you for the lightning fast follow up!

PR #691, opened 2024-09-16
[Bug]: Check if User is allowed to create a class when pasting

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/690


kingjia90:
For documents, unfortunately is not applicable because it works on the permission of the document type (these "presets/boilerplate")
image
but once you created a document, it lose the info from which preset it was created and it could be overidden anyways.

image

fashxp:
> For documents, unfortunately is not applicable because it works on the permission of the document type (these "presets/boilerplate")

Ahh, true ... Then just for data objects it is 👍

PR #692, opened 2024-09-16
[Grid] Support sorting by object brick quantity value field

BlackbitDevs:
Steps to reproduce:
1. Create class `Car` with object brick field `attributes`
2. Create object brick `Bodywork`, assign `Car` as allowed class and add quantity value field `cargoCapacity`
3. Go to grid for `Home` (aka `/`), add grid column `Bodywork.cargoCapacity` and sort by `cargoCapacity`

Result:
SQL error 
`SQLSTATE[42S22]: Column not found: 1054 Unknown column 'Bodywork.cargoCapacity' in 'order clause'`

This PR fixes this similar as it is done in https://github.com/pimcore/admin-ui-classic-bundle/blob/d33a87cd8c8cdc5ef203cb83ea49f781a30fb77b/src/Helper/GridHelperService.php#L564-L570 for class fields of type quantity value (and RgbaColor).

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -566,19 +566,20 @@ class GridHelperService
                     $orderKey = 'concat(' . $orderKey . '__rgb, ' . $orderKey . '__a)';
                     $doNotQuote = true;
                 } elseif (strpos($orderKey, '~') !== false) {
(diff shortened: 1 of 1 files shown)


mcop1:
Thanks :+1:

Issue #688, opened 2024-09-13
Cell doesn't take full width in table type


mhgtlc:
That issue appeared after version `1.5.2` but not sure in which exactly
<img width="1498" alt="Screenshot 2024-09-13 at 09 14 00" src="https://github.com/user-attachments/assets/22e58f40-9c0f-45f5-8168-855102bea062">

PR #694, opened 2024-09-17
Fixes cell width not taking all space

mhgtlc:
Fixes https://github.com/pimcore/admin-ui-classic-bundle/issues/688

### DIFF
--- a/public/js/pimcore/object/tags/table.js
+++ b/public/js/pimcore/object/tags/table.js
@@ -225,16 +225,7 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
             tbar: tbar,
             viewConfig: {
                 markDirty: false,
-                forceFit: true,
-                listeners: {
-                    refresh: function (dataview) {
-                        Ext.suspendLayouts();
-                        Ext.each(dataview.panel.columns, function (column) {
-                            column.autoSize();
-                        });
-                        Ext.resumeLayouts(true);
-                    }
-                }
+                forceFit: true
             }
         });
         this.component.add(this.grid);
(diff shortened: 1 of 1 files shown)


robertSt7:
@mhgtlc Thanks for coming up with a solution. But I think when we remove this code then we will have this problem again, right? 
https://github.com/pimcore/admin-ui-classic-bundle/pull/566

mhgtlc:
> @mhgtlc Thanks for coming up with a solution. But I think when we remove this code then we will have this problem again, right? #566

@robertSt7 probably yes, in this case we need another solution I guess

public/js/pimcore/object/tags/table.js
```diff
@@ -225,16 +225,7 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
...
-                listeners: {
-                    refresh: function (dataview) {
-                        Ext.suspendLayouts();
-                        Ext.each(dataview.panel.columns, function (column) {
-                            column.autoSize();
-                        });
-                        Ext.resumeLayouts(true);
-                    }
-                }
+                forceFit: true
```
kingjia90:
```suggestion
                forceFit: true,
                listeners: {
                    refresh: function (dataview) {
                        Ext.suspendLayouts();
                        let totalWidth = 0;
                        const columns = dataview.panel.columns;
                        
                        Ext.each(columns, function (column) {
                            totalWidth += column.getWidth();
                        });
                        
                        if (totalWidth > this.fieldConfig.width){
                            Ext.each(columns, function (column) {
                                column.autoSize();
                            });
                        }

                        Ext.resumeLayouts(true);
                    }.bind(this)
                }
```
This would be the best of the 2 worlds, so when the column is not occupying all the width, it would try to spread, while when is too crowded, it would use the autofit and display scrollbar

wdyt?

kingjia90: Without listener (as the change in this PR)
image

With listener (https://github.com/pimcore/admin-ui-classic-bundle/pull/566)
image


with the suggestion above
image

kingjia90 (changes requested):

PR #702, opened 2024-09-26
fix: tooltip error for not editable many to many relation

kjkooistra-youwe:
Error: application style and tooltip aborted, nor matching element found.
Caused by missing object_field class, which is used by edit.js to find the relevant element to apply the tooltip to.
Fixed by adding class(es) to component in getLayoutShow.


youwe-petervanderwal:
@kjkooistra-youwe  Bugfixes should go into the 1.5 branch

kjkooistra-youwe:
I can't seem to modify the target branch, so I'll close it and create a new one.

PR #704, opened 2024-09-27
fix: tooltip error for not editable many to many relation

kjkooistra-youwe:
Error: application style and tooltip aborted, nor matching element found.
Caused by missing object_field class, which is used by edit.js to find the relevant element to apply the tooltip to.
Fixed by adding class(es) to component in getLayoutShow.


mattamon (approved):
Thanks for your contribution, works like a charm!

PR #705, opened 2024-09-27
Fix: Asset metadata of type object in assets gridview is not updated when value changes

code-status-404:
Custom asset metadata of type object cant be set in asset gridview.

Steps to reproduce:
- Creating a custom asset meta data of type object and assign it to an image
- Switch to grid view of the asset directory
- Selecting the created custom metadata of type object for display in the grid
- Click on the field and link a data object
- In the CellEditor click on save 
=> The relation is not adopted

The change is not recognized as a change in the manyToOneRelation class because the variable "this.dataChanged" never changes to "true".

Error was introduced by a change from December 4th, 2023 - commit e422dca4 in js class "pimcore.element.helpers.gridCellEditor" line 123 

Found in Pimcore 11.3.2 with admin-ui-classic-bundle version 1.5.4


code-status-404:

code-status-404:
Sorry, the wrong target branch is slected, where can i change it?

fashxp:
Hmm but for bugfixes the target branchnis correct... You would need to rebase your branch and force push

PR #710, opened 2024-10-02
Fix undefined error when copy and paste

torqdev:
When copying and pasting inside the DataObject editor, it throws an undefined error in the console when a reverse relation exists.


mattamon (approved):
Awesome thank you! PRs like these are much appreciated and increase the overall stability of the system. :)

PR #715, opened 2024-10-08
Field type Table: Copy / Paste

BlackbitDevs:
This PR makes the paste dialog for `table` fields more usable: Before, the paste window immediately applies the value and closes the dialog window after pasting. This means you cannot review the values. This does not only happen on pasting but on any changing of the value (so if you enter "a" (any character) in the paste dialog text area, this value also gets applied immediately).
As @mcop1 suggested in https://github.com/pimcore/admin-ui-classic-bundle/pull/559#issuecomment-2277401497 this PR adds a "Save" button to the paste dialog so that you can paste the content to the text area, review it and apply by clicking the "Save" button:
Bildschirmfoto 2024-10-08 um 15 38 00


mcop1:
Aaaah nice, thank you :+1:

PR #719, opened 2024-10-11
Improvement: move annotations to attributes

michalananapps:
Migrating annotations from Symfony\Component\Routing\Annotation\Route to attributes has several advantages, which can make project cleaner, more modern, and maintainable. Migrating to attributes is needed to move to newer versions of Symfony in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,8 +23,8 @@
     "codeception/module-asserts": "^2.0.1",
     "codeception/module-symfony": "^3.1.1",
     "codeception/phpunit-wrapper": "^9",
-    "phpstan/phpstan": "1.10.30",
-    "phpstan/phpstan-symfony": "^1.3.2",
+    "phpstan/phpstan": "^1.12",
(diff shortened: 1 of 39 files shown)


lukmzig:
@fashxp was there a change in the decision regarding the migration of annotations to attributes? We had a similar PR some time ago and we decided that we would not proceed there, see comment: 
https://github.com/pimcore/admin-ui-classic-bundle/pull/500#issuecomment-2315500690

fashxp:
@lukmzig you are right. No other decision or change of priorities. We should not merge this due to our priorities (focus on studio and reducing maintenance efforts for admin-ui-classic to a minimum as it will fade out as soon as studio is ready). 

@michalananapps I'm really sorry :face_with_diagonal_mouth:, thank you very much for your efforts though!!

lukmzig:
@fashxp thank you for the clarification.

@michalananapps thank you for your efforts! I am sorry for this situation but as mentioned we will close this PR in favor of the studio

PR #724, opened 2024-10-17
[Edit lock] Check for edit lock when editing data, not when opening element

BlackbitDevs:
Currently when 2 users open the same element, there is this warning:
Bildschirmfoto 2024-10-17 um 15 56 08

This confuses users because this even happens if they only want to look up things without changing anything. For this reason this PR changes the behaviour so that the edit lock does not get checked when *opening* an element but when *editing* it. As soon as any data of the element gets changed and thus the element gets `dirty`, the edit lock will get checked. The message in the popup window has been adjusted a bit:
Bildschirmfoto 2024-10-17 um 16 04 00

When the user chooses to continue editing (= clicks "Yes"), the edit lock will get released.
When the user clicks "No", the element will get closed and reopened.


BohdanSheyko:

mattamon:
@fashxp how will it be done in the new studio ui? Conceptual it sounds good to me.

fashxp:
not finally decided yet. but we have a concept to get notified about remote changes and what you want to do about them. so that edit lock might not be necessary at all anymore. 

this change needs to be tested carefully though. maybe someone from the community can help us there?

mcop1:
Hello @BlackbitDevs ,
thanks for the pr,  but I got the following error after checking out the branch: 
`Attempted to call an undefined method named "getCachedSymfonyEnvironments" of class "Pimcore\Tool"`

It indicates that the branch isnt up to date, could you please update it and have a look at the merge conflict?

astapc:
@BlackbitDevs Could you please respond within the next week?

public/js/pimcore/element/abstract.js
```diff
@@ -194,7 +195,39 @@ pimcore.element.abstract = Class.create({
...
+                Ext.Ajax.request({
+                    url: Routing.generate('pimcore_admin_element_islocked'),
+                    method: 'GET',
+                    async: false,
+                    params: {
+                        id: this.id,
+                        type: elementType
+                    },
+                    success: function (response) {
+                        var result = Ext.decode(response.responseText);
```
kingjia90:
```suggestion
                        const result = Ext.decode(response.responseText);
```

src/Controller/Admin/ElementController.php
```diff
@@ -37,7 +38,23 @@
  */
 class ElementController extends AdminAbstractController
 {
+    use ElementEditLockHelperTrait;
+
     #[Route('/element/lock-element', name: 'pimcore_admin_element_lockelement', methods: ['PUT'])]
+    public function isLockedAction(Request $request): JsonResponse
+    {
+        $isLocked = Element\Editlock::isLocked($request->query->getInt('id'), $request->query->get('type'), $request->getSession()->getId());
```
kingjia90:
```suggestion
        $isLocked = Element\Editlock::isLocked(
            $request->query->getInt('id'), 
            $request->query->get('type'), $request->getSession()->getId()
        );
```

kingjia90: omg i was sleepy, improved again by https://github.com/pimcore/admin-ui-classic-bundle/pull/724/commits/8e65d84e87e9d74b3be33be0de64308aea08a6d6 😄

kingjia90:
if i got correctly

Before PR
User 1 is just viewing, user 2 gets the warning even if both are just viewing
User 1 is editing, user 2 gets the warning on open/view

After PR
User 1 is just viewing, user 2 is also viewing without any plan to make changes, no warnings at all (which is good)
User 1 is editing, user 2 doesn't get a warning (even if is just viewing) thus is not gonna be aware that the data could be outdated anytime soon (potentially bad?)
User 1 is editing, user 2 tries to apply some changes, it warns to avoid overriding what user 1 was doing

mcop1:
> User 1 is editing, user 2 doesn't get a warning (even if is just viewing) thus is not gonna be aware that the data could be outdated anytime soon (potentially bad?)

With the current behavior it is only slightly (if even) better in my opinion. Yes, you get the dialog when someone else has opened the property dialog, but that doesnt mean that someone is editing it. Then, if you proceed and click on "Yes" you wont get another notification if someone is actually starting to edit it or even saving a new version. 
We cant solve that in a satisfying way until we have some other technology in place like we probably have in StudioUI. Until then, in my opinion, the solution in the pr is the better one.

kingjia90:
Yes, i am also in favor of this PR, anything that reduces popups has my vote 😄  

Wanted to put down a recap of the behavior change, might be that we need to make these two approach as system-wide configurable option.


I was also thinking that we needed something like other collaborative platforms like Google Docs and Jira Confluence
image where the users actively viewing/editing on something appear in real time on the toolbar but might not be worth to add it to the sunsetting admin ui.

mcop1:
> Yes, i am also in favor of this PR, anything that reduces popups has my vote 😄
> 
> Wanted to put down a recap of the behavior change, might be that we need to make these two approach as system-wide configurable option.
> 
> I was also thinking that we needed something like other collaborative platforms like Google Docs and Jira Confluence image where the users actively viewing/editing on something appear in real time on the toolbar but might not be worth to add it to the sunsetting admin ui.

Agreed, we should look for a better solution in studio. Think the google docs approach is kinda nice. 
So this pr is ready to merge? 

EDIT: Thanks for providing the pr :+1:

mcop1 (approved):

PR #725, opened 2024-10-17
Fix Asset exists check for files within folders when drag and dropping.

### LINKED ISSUE #626: Drag and Drop Local Folder of Files on Tree View Does not Prompt for Overwrite
Currently If I drop a _folder_ of local files onto the tree view I am not prompted to "overwrite" vs "keep both". It just defaults to keep both. Conversely, if I drop one more _files_ onto the tree view, I am prompted  "overwrite" vs "keep both".

Related #420 where uploading zip files does not currently prompt.

cancan101:
Closes #626

The asset is not correctly resolved in the exists check if it is in a folder. 

Currently, the `add-asset` endpoint is passed the directory path but the `exists` endpoint is not.

### DIFF
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -400,15 +400,16 @@ class AssetController extends ElementControllerBase implements KernelControllerE
     public function existsAction(Request $request): JsonResponse
     {
         $parentAsset = \Pimcore\Model\Asset::getById((int)$request->get('parentId'));
-        if ($dir = $request->get('dir')){
+
+        $dir = '';
+        if ($request->get('dir')){
+            $dir = $request->get('dir');
             // this is for uploading folders with Drag&Drop
             // param "dir" contains the relative path of the file
             if (strpos($dir, '..') !== false) {
                 throw new \Exception('not allowed');
             }
             $dir =  '/' . trim($dir, '/ ');
-        } else {
-            $dir = '';
         }
 
         $assetPath = $parentAsset->getRealFullPath() . $dir . '/' . $request->get('filename');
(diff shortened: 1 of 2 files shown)


src/Controller/Admin/Asset/AssetController.php
```diff
@@ -398,9 +398,21 @@ public function addAssetCompatibilityAction(Request $request, Config $config): J
...
         $parentAsset = \Pimcore\Model\Asset::getById((int)$request->get('parentId'));
+        if ($dir = $request->get('dir')){
+            // this is for uploading folders with Drag&Drop
+            // param "dir" contains the relative path of the file
+            if (strpos($dir, '..') !== false) {
+                throw new \Exception('not allowed');
+            }
+            $dir =  '/' . trim($dir, '/ ');
+        } else {
+            $dir = '';
```
mattamon: I guess we could remove the else part and make the dir empty by default?

cancan101: `$dir` is set above from the request param so even if it was initialized to empty string it would be re-set to null. or did you mean something else?

mattamon: Ah sorry, seems I missed that it was not equal but assigning in the if statement.

I personally dislike these kind of if statements since they only work in PHP. 

I'd rather get the `$dir` first from the `$request` and then checking if something is in the request.
At least for me it would enhance the readability.

E.g.
```php
// Initialize "dir" with an empty string by default
$dir = '';

// Get the "dir" parameter from the request, if available
if ($request->get('dir')) {
    $dir = $request->get('dir');
    ...
}
```

blankse: Why not use the default argument of the `get` function?
```
$dir = $request->query->get('dir', '');

if ($dir) {
    ...
}
```

cancan101:
any other comments on the PR?

mattamon:
> any other comments on the PR?

Hey @cancan101,
since it is a bugfix it should go in the bugfix branch no?

cancan101:
done

mattamon (approved):
Thank you for the PR. Sorry that it took a while!

PR #731, opened 2024-10-23
BUG: display the nice path if field type combo is not editable

patrickbitzer:
I had the phenomenon that when I had set a OneToManyRelation field with the Inline Search type, the PathFormatter was not applied there. The problem occurres only when the user is in not editable mode.

image

I have made an adjustment that fixes the problem.

image-20241022-123834

### DIFF
--- a/public/js/pimcore/object/tags/manyToOneRelation.js
+++ b/public/js/pimcore/object/tags/manyToOneRelation.js
@@ -671,8 +671,9 @@ pimcore.object.tags.manyToOneRelation = Class.create(pimcore.object.tags.abstrac
         if (!this.object) {
             return;
         }
+
         let targets, responseHandler;
-        if (pimcore.helpers.hasSearchImplementation() && this.fieldConfig.displayMode === 'combo') {
+        if (pimcore.helpers.hasSearchImplementation() && this.fieldConfig.displayMode === 'combo' && !this.fieldConfig.noteditable) {
             targets = this.store.data;
             responseHandler = function (responseData) {
                 this.component.removeCls('grid_nicepath_requested');


patrickbitzer:

patrickbitzer:
I updated the description

blankse:
@patrickbitzer Target branch for bug fixes should be `1.6`. Can you rebase it?

patrickbitzer:
@blankse maybe done.

blankse:
@patrickbitzer You still have commits from 1.x in your branch. You need to rebase with the `--onto` option to remove them:
`git rebase --onto origin/1.6 origin/1.x`
Then, you need to push with the `--force` option:
`git push --force`

patrickbitzer:
@blankse sorry. I have done now

mattamon:
@patrickbitzer Could you explain this a little more? What do you mean by inline search type?
I cannot reproduce this with my local instance.

blankse (approved):

blankse:
@mattamon I could reproduce it:

1. Create a Many-To-One Relation field with Formatter Service and Display Mode "Inline Search"
2. Create an object and fill the field => The formatter is used correctly
3. Change the field definition to "Not editable"
4. Reload the object => The formatter is not used and you see the path of the relation object

IronSean:
@mattamon we're also experiencing this with our clients.

blankse:
@mattamon Still can't reproduce it?

mattamon:
@blankse sorry I am no longer working on this.  @robertSt7 can you please take over.

robertSt7:
@patrickbitzer Thanks for fixing displaying of the path

PR #734, opened 2024-10-28
Fix background for external transparent image

mhgtlc:
For external image:
Screenshot 2024-10-28 at 12 06 26

when uploading a transparent image, it looks like:
Screenshot 2024-10-28 at 12 06 00


mattamon (approved):
Awesome! Thanks for the fix!

PR #739, opened 2024-11-01
Update PimcoreApp-all_1.css

djbodya:
This fix must remove error like so "No route found https://pimcore.mamadecor.pl/bundles/pimcoreadmin/extjs/images/grid/col-move-top.png%22%20(from%20%22https://pimcore.mamadecor.pl/bundles/pimcoreadmin/extjs/css/PimcoreApp-all_1.css"

### DIFF
--- a/public/extjs/css/PimcoreApp-all_1.css
+++ b/public/extjs/css/PimcoreApp-all_1.css
@@ -13308,7 +13308,7 @@ a.x-color-picker-item:hover {
     position: absolute;
     top: -8px;
     left: -12px;
-    background-image: url(../images/grid/dd-insert-arrow-right.png);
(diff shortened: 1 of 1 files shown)


mattamon:
Could you please explain a little more in detail what you actually mean?
I can see the changes in the directory names, but I would need a more detailed description. Is there an existing issue for this problem?

djbodya:
An error appears on my open-source installation. 
The path which is given in this file https://pimcore.mamadecor.pl/bundles/pimcoreadmin/extjs/css/PimcoreApp-all_1.css does not link to the image file because this path does not exist. 
I updated the right path in this merge request. It should fix the problem.
There is no created issue.

mattamon (approved):
Thanks for your contribution!

PR #740, opened 2024-11-04
[2FA] Set default issuer to "Pimcore" instead of "Pimcore 2 Factor Authentication"

BlackbitDevs:
Currently the QR code for 2FA is set to "Pimcore 2 Factor Authentication" by default. This is not optimal because:
1. All applications in Google Authenticator app are 2 factor authentications, so it does not make sense to write this information
2. The string "Pimcore 2 Factor Authentication" is so long that the actual project name is not visible on most mobile phones (in portrait mode) - even if you rename the application in Google Authenticator, the issuer name is kept. Thus if you have multiple Pimcore systems in Authenticator app, you always have to use landscape mode to see which Pimcore app to use

With this PR, the issuer name gets changed to "Pimcore". So by default the app in Google Authenticator would get named

> Pimcore: username@Pimcore

If you change the app name in Authenticator, you will still have

> Pimcore: App name

### DIFF
--- a/config/pimcore/two_factor_authentication.yaml
+++ b/config/pimcore/two_factor_authentication.yaml
@@ -3,7 +3,7 @@ scheb_two_factor:
     google:
         enabled: true                                       # If Google Authenticator should be enabled, default false
         server_name: Pimcore                                # Server name used in QR code
-        issuer: Pimcore 2 Factor Authentication             # Issuer name used in QR code
+        issuer: Pimcore                                     # Issuer name used in QR code
 
     security_tokens:
         - Pimcore\Bundle\AdminBundle\Security\Authentication\Token\TwoFactorRequiredToken


kingjia90:
Thank you, agree, in addition, seeing a `2` after `Pimcore` mislead me for Pimcore v2.0 😄 a literal `two` would have been better.

The question now is, should it be in a bugfix branch as `Task` or is something should be in a minor? Shouldn't be bc-breaking, right?

jdreesen:
I think it's just the name of the issuer in the QR code. So it is only used when 2FA is set up by a user. The value can be changed by the user in their 2FA app anyway. So I don't think this is breaking.

BlackbitDevs:
Exactly, this is not a BC break.

@jdreesen 
> The value can be changed by the user in their 2FA app anyway.

Sadly, this is not completely true. At least in Google Authenticator when you rename an item, you can only rename the `server_name` part but not the `issuer` part from 
https://github.com/pimcore/admin-ui-classic-bundle/blob/e3b66fb164a72430cee9bac9ccf1ff21b6de9d3d/config/pimcore/two_factor_authentication.yaml#L5-L6
(This is what it makes it so annoying because in portrait mode, I only see loads of "Pimcore 2 Factor Authentication: " and then the first 3 characters which I can rename) - and always have tu turn my phone to landscape mode to see the full names.

robertSt7:
@BlackbitDevs Thanks for the improvement

PR #741, opened 2024-11-05
Fix: Filter for relation throws an error

### LINKED ISSUE #720: [Bug] Filter for relation throws an error
The relation filters seem to be broken since version 1.6.1.

When trying to filter for relation ExtJS is throwing an error

Bildschirmfoto 2024-10-14 um 15 33 33

blankse:
Fix: https://github.com/pimcore/admin-ui-classic-bundle/issues/720

### DIFF
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -449,8 +449,8 @@ pimcore.element.helpers.gridColumnConfig = {
 
             const title = t("filter_by_relation_field") + " " + fieldInfo.text;
             let width = 700;
-            if (tagType === 'manyToManyObjectRelation' && fieldInfo.layout.layout.width) {
-                width = fieldInfo.layout.layout.width + 25;
+            if (tagType === 'manyToManyObjectRelation' && fieldInfo.layout.layout.width && fieldInfo.layout.layout.width !== '100%') {
+                width = sumWidths(fieldInfo.layout.layout.width, 25);
             }
             this.filterByRelationWindow = new Ext.Window({
                 autoScroll: true,
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -386,9 +386,9 @@ pimcore.element.helpers.gridColumnConfig = {
     },
 
     filterPrepare: function (column) {
-        var dataIndexName = column.dataIndex
-        var gridColumns = this.grid.getColumns();
-        var columnIndex = -1;
+        const dataIndexName = column.dataIndex
+        const gridColumns = this.grid.getColumns();
+        let columnIndex = -1;
         for (let i = 0; i < gridColumns.length; i++) {
             let dataIndex = gridColumns[i].dataIndex;
             if (dataIndex == dataIndexName) {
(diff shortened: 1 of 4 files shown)


public/js/pimcore/element/helpers/gridColumnConfig.js
```diff
@@ -456,7 +456,7 @@ pimcore.element.helpers.gridColumnConfig = {
             title: title,
             items: [formPanel],
             bodyStyle: "background: #fff;",
-            width: formPanel.items.items[0].width + 25,
+            width: 650,
```
kingjia90: Thank you for the PR!

LGMT but it removes the possibility to define a custom width(?).
Haven't tested it yet, but what about something like 
```suggestion
            width: (formPanel.items?.items[0]?.width ?? 625) + 25,
```
?

blankse: @kingjia90 `formPanel.items.items[0].width` is undefined. Regardless of whether a width is set or not.

kingjia90: Ah ok, then wondering why it was changed to this in https://github.com/pimcore/admin-ui-classic-bundle/pull/681  
@mattamon could you please have a second look on this?

blankse: @kingjia90 Ah, there is a many to many field used. I will check this. I only test it with many to one.

blankse: @kingjia90 I changed it.

kingjia90:
There is still something not right, maybe we should introduce also a minimum width? 
image

blankse:
@kingjia90 What did you config as width?

kingjia90:
it's the default value in demo instance, ~i believe it's empty~ it's 96 but don't know from where it comes
image

blankse:
@kingjia90 I changed it again https://github.com/pimcore/admin-ui-classic-bundle/pull/741/commits/1bddd1241a6652cd2c896c030eed13f5919cec17
There are so many width keys in this fieldInfo object. I hope this is the right one. In my tests it works now.

kingjia90:
weird, to me it doesn't work, it's the default `700`, the table within the dialog seems crammed and not utilizing/distributing the width for each grid column

image

blankse:
@kingjia90 What did you set as the width? I entered 1000 in this field in the demo, and it is now wider. Yes, the columns are narrow, but I believe that’s because they are empty at the beginning.

blankse:
@kingjia90 I cleared the browser cache/cookies. Now the column widths also correct. Extjs saves the widths of the columns in the local storage:
image
image

kingjia90:
> @kingjia90 What did you set as the width? I entered 1000 in this field in the demo, and it is now wider. Yes, the columns are narrow, but I believe that’s because they are empty at the beginning.

i didn't set anything, that should be right, nevermind, it is just fine


> @kingjia90 I cleared the browser cache/cookies. Now the column widths also correct. Extjs saves the widths of the columns in the local storage: image image

ahhh didn't know that... does it implicitly mean that we cannot use variable width and should be always fixed?

blankse:
@kingjia90 It's not from ext js. It is here in the admin-ui-classic-bundle:

https://github.com/pimcore/admin-ui-classic-bundle/blob/e3b66fb164a72430cee9bac9ccf1ff21b6de9d3d/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L405-L426
https://github.com/pimcore/admin-ui-classic-bundle/blob/e3b66fb164a72430cee9bac9ccf1ff21b6de9d3d/public/js/pimcore/object/tags/abstractRelations.js#L210-L225

You have first open the filter with the very small window. So the columns was saved small. You can drag and drop the widths wider and it should be saved for the next window, right? When you have a empty local storage the columns have a correct width.

kingjia90:
It seems it's because the grid is likely empty on first time opening as "Filter by relation" and therefore it doesn't have an estimation on how long each column would be. Ideally we should trigger this resize with `localStorage.setItem` on selecting/adding the first element

The current PR changes LGTM

kingjia90 (approved):

PR #748, opened 2024-11-14
[Improvement] Refactor query for performance gain

### LINKED ISSUE #700: GridView Performance Degradation with Multiple Roles and Workspaces
### Steps to reproduce

1. Create a user
2. Create at least 5 roles and assign multiple workspaces to each role
3. Build a folder structure with several subfolders
4. Add a large number of objects to the folder structure
5. Open the GridView
6. A significant slowdown should be noticeable

#### Blackfire Results
Roles: 8
Workspaces: 50 (partially duplicated)
grafik

As an admin, the process is much faster since workspace queries are not included: 
grafik

### Actual Behavior
The loading times are extremely long and can reach up to 20 seconds or more, depending on the configuration, often leading to timeouts that make the GridView unusable.

### Expected Behavior
Faster SQL queries to significantly reduce loading times, avoid timeouts and ensure that the GridView remains usable without long waiting times.

mattamon:
Resolves #700 - at least it is a start.

Refactoring the query is one part. 
We should also introduce an index like the following, but we can do this in the core itself with a migration.
See https://github.com/pimcore/pimcore/pull/17834

```sql
CREATE INDEX idx_users_workspaces ON users_workspaces_object (userId, cpath, list);
CREATE INDEX idx_users_workspaces ON users_workspaces_asset (userId, cpath, list);
```

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -616,18 +616,12 @@ class GridHelperService
         if (!$adminUser->isAdmin()) {
             $userIds = $adminUser->getRoles();
             $userIds[] = $adminUser->getId();
-            $conditionFilters[] = '(
-                    EXISTS (
-                        SELECT 1
-                        FROM users_workspaces_object
-                        WHERE userId IN (' . implode(',', $userIds) . ')
-                        AND (
-                            LOCATE(CONCAT(`path`, `key`), cpath) = 1
-                            OR LOCATE(cpath, CONCAT(`path`, `key`)) = 1
-                        )
-                        AND list = 1
-                    )
-                )';
+            $ids = implode(',', $userIds);
+            $conditionFilters[] = " (
+                (select list from users_workspaces_object where userId in ($ids) AND CONCAT(`path`, `key`) LIKE CONCAT(cpath, '%')  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
+                OR
+                (select list from users_workspaces_object where userId in ($ids) AND cpath LIKE CONCAT(`path`, `key`, '%')  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
+             )";
         }
 
         $featureJoins = [];
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -616,11 +616,18 @@ class GridHelperService
         if (!$adminUser->isAdmin()) {
             $userIds = $adminUser->getRoles();
             $userIds[] = $adminUser->getId();
-            $conditionFilters[] = ' (
-                                                    (select list from users_workspaces_object where userId in (' . implode(',', $userIds) . ') and LOCATE(CONCAT(`path`,`key`),cpath)=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
-                                                    OR
-                                                    (select list from users_workspaces_object where userId in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(`path`,`key`))=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
-                                                 )';
+            $conditionFilters[] = '(
+                    EXISTS (
+                        SELECT 1
+                        FROM users_workspaces_object
+                        WHERE userId IN (' . implode(',', $userIds) . ')
+                        AND (
+                            LOCATE(CONCAT(`path`, `key`), cpath) = 1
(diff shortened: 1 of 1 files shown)


fashxp:
Sadly, this query does not deliver the correct results :disappointed:.

Take the following permission configuration which must not show `/Product Data/Accessories/rims`: 
image

The old query looks something like this for userid 20
```sql
SELECT id, path, `key` FROM objects WHERE path = '/Product Data/Accessories/' AND `key` = 'rims' AND
((select list from users_workspaces_object where userId in (20) and LOCATE(CONCAT("/Product Data/Accessories/","rims"),cpath)=1
    ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
OR
(select list from users_workspaces_object where userId in (20) and LOCATE(cpath,CONCAT("/Product Data/Accessories/","rims"))=1
    ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
); 
```
... and returns zero results - which is correct. 

The new query looks something like this 
```sql
SELECT id, path, `key` FROM objects WHERE path = '/Product Data/Accessories/' AND `key` = 'rims' AND
    EXISTS (
        SELECT 1
        FROM users_workspaces_object
        WHERE userId IN (20)
          AND (
            LOCATE(CONCAT("/Product Data/Accessories/","rims"), cpath) = 1
                OR LOCATE(cpath, CONCAT("/Product Data/Accessories/","rims")) = 1
            )
          AND list = 1
    );
``` 
... and returns results - which is wrong. 


The reason why while testing in the UI it doesn't show a result is [that line](https://github.com/pimcore/admin-ui-classic-bundle/blob/e2f0b7723762c655190844c805df897c607745bb/src/Controller/Admin/DataObject/DataObjectActionsTrait.php#L128) - which double checks permissions with calling `isAllowed` for each element. 

This in turns becomes a problem for example with following permission configuration, which should show one line in the grid:
image

With the old query, it works as expected. 
With the new query it doesn't - because the query delivers a whole page of false-positives (which are filtered out) so the true-positive doesn't make it into the result.

mattamon:
> Sadly, this query does not deliver the correct results 😞.
> 
> Take the following permission configuration which must not show `/Product Data/Accessories/rims`: image
> 
> The old query looks something like this for userid 20
> 
> ```sql
> SELECT id, path, `key` FROM objects WHERE path = '/Product Data/Accessories/' AND `key` = 'rims' AND
> ((select list from users_workspaces_object where userId in (20) and LOCATE(CONCAT("/Product Data/Accessories/","rims"),cpath)=1
>     ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
> OR
> (select list from users_workspaces_object where userId in (20) and LOCATE(cpath,CONCAT("/Product Data/Accessories/","rims"))=1
>     ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
> ); 
> ```
> 
> ... and returns zero results - which is correct.
> 
> The new query looks something like this
> 
> ```sql
> SELECT id, path, `key` FROM objects WHERE path = '/Product Data/Accessories/' AND `key` = 'rims' AND
>     EXISTS (
>         SELECT 1
>         FROM users_workspaces_object
>         WHERE userId IN (20)
>           AND (
>             LOCATE(CONCAT("/Product Data/Accessories/","rims"), cpath) = 1
>                 OR LOCATE(cpath, CONCAT("/Product Data/Accessories/","rims")) = 1
>             )
>           AND list = 1
>     );
> ```
> 
> ... and returns results - which is wrong.
> 
> The reason why while testing in the UI it doesn't show a result is [that line](https://github.com/pimcore/admin-ui-classic-bundle/blob/e2f0b7723762c655190844c805df897c607745bb/src/Controller/Admin/DataObject/DataObjectActionsTrait.php#L128) - which double checks permissions with calling `isAllowed` for each element.
> 
> This in turns becomes a problem for example with following permission configuration, which should show one line in the grid: image
> 
> With the old query, it works as expected. With the new query it doesn't - because the query delivers a whole page of false-positives (which are filtered out) so the true-positive doesn't make it into the result.

Valid points. I used now only some micro-optimization since LIKE seems to performe better than Locate.

At least the results look now better to me but I am not sure what big of an impact it has. 
For my tests we are not far off the admin time.

cancan101:
@mattamon @kingjia90 was https://github.com/pimcore/pimcore/pull/17834 ever re-PRed against a feature branch?

mattamon:
@cancan101 yes :) https://github.com/pimcore/pimcore/pull/17847

PR #757, opened 2024-11-19
[Bug] Readding check to solve update/move problem of objects

### LINKED ISSUE #756: [Bug] Moving object directly onto folder without the folder being loaded resulted in the object never getting updated and moved
Based on this PR https://github.com/pimcore/admin-ui-classic-bundle/pull/469/files#diff-3b41334b430020814d1a30721473a6d8647e88045cb572be36931f80d28522f2L277 where a check was removed, the update function was never called, thus not updating or moving the object.

mattamon:
Resolves #756

### DIFF
--- a/public/js/pimcore/object/tree.js
+++ b/public/js/pimcore/object/tree.js
@@ -364,7 +364,7 @@ pimcore.registerNS("pimcore.object.tree");
          }
 
          // dropping objects not allowed if the tree/folder is paginated and sort by index (manual indexes) is enabled
-         if(((!newParent.pagingData.canSortManually) || (newParent.childNodes.length > pimcore.settings['object_tree_paging_limit'])) && (newParent.data.sortBy == "index")){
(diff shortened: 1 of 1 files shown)


public/js/pimcore/object/tree.js
```diff
@@ -276,7 +276,9 @@ pimcore.registerNS("pimcore.object.tree");
                      this.onTreeNodeMove(record, record.parentNode, overModel, 0);
                  }
              }.bind(this));
+         }
 
+         if (typeof this.treeNodeMoveParameter.oldParent.getOwnerTree !== "function") {
```
kingjia90, line 282:
```suggestion
         // recheck wheter the getOwnerTree is a function after reassignment onTreeNodeMove()
         if (typeof this.treeNodeMoveParameter.oldParent.getOwnerTree !== "function") {
```
it should be helping to spot that the getOwnerTree could be changed by `onTreeNodeMove()`
https://github.com/pimcore/admin-ui-classic-bundle/pull/757/files#diff-3b41334b430020814d1a30721473a6d8647e88045cb572be36931f80d28522f2R263

kingjia90 (commented):
just suggesting to add a comment to explain why it has to be there even though it looks like repetead code

PR #750, opened 2024-11-15
[Improvement] Add option editor to select options

kjkooistra-youwe:
Allow editing options as comma separated values.
Refactored option editor to allow using custom column names.

Relates to
https://github.com/pimcore/admin-ui-classic-bundle/pull/291
https://github.com/pimcore/pimcore/pull/14454

PR #759, opened 2024-11-21
[Translations] Incorrect translation for separated

kjkooistra-youwe:
Relates to https://github.com/pimcore/admin-ui-classic-bundle/pull/750

I noticed the Pimcore core bundle also still contains a few of these translations.
https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/translations/admin.sk.yaml#L526
https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/translations/admin.sv_fi.yaml#L516


robertSt7:
Hi @kjkooistra-youwe  thanks for the contribution.

Since this looks more like a bugfix, could you please re-base your PR to 1.7?

I also see that you change some non-English translations files.
Actually only the update of the English translations have an effect, because these keys are then synced with POEditor. https://poeditor.com/join/project/XliCYYgILb
You can then translate it there to other languages.

Could you please remove all changes from the non-English files?

Thank you

kjkooistra-youwe:
@robertSt7 Sure. I figured this might work, since only the keys were updated and not the actual translations. Does that mean all translations will have to be added again manually? I'm not familiar with poeditor.

brusch:
@kjkooistra-youwe yes, all languages other than EN are managed in POEditor and must be entered there. However the keys are getting imported from the English translations from the repository automatically to POEditor.

robertSt7:
@kjkooistra-youwe Thanks for the PR

PR #763, opened 2024-11-21
[Relational fields] Sort options for column filters alphabetically

BlackbitDevs:
Follow-up of #659

Currently the order of column filter options for (advanced) many-to-many (object) relation fields depends on the order of assigned elements in the relation. This makes it hard to find a certain filter value. This PR sorts the filter values alphabetically.

Before:
Bildschirmfoto 2024-11-21 um 19 05 03
(Imagine a relation with 100 items, then finding the desired value will take some time)

After:
Bildschirmfoto 2024-11-21 um 19 04 16


robertSt7:
@BlackbitDevs Thanks for the sorting of the filter options

PR #764, opened 2024-11-22
[Task] Add endroid dependency

mattamon:
As discussed here: https://github.com/pimcore/pimcore/pull/17799#issuecomment-2477236052


mattamon:
Closed in favor of general dependency upgrade check

PR #765, opened 2024-11-22
[Select Options] Admin only restriction

kjkooistra-youwe:
Only allow admin users to update or delete select option definition.

Relates to https://github.com/pimcore/pimcore/pull/17861


robertSt7:
@kjkooistra-youwe Thanks for this permission improvement

PR #767, opened 2024-11-25
[Bug]: Fix GridView Performance and results

### LINKED ISSUE #700: GridView Performance Degradation with Multiple Roles and Workspaces

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/700

## Additional Info
I'd would just fix it like this, it is the exact [well covered and tested ](https://github.com/pimcore/pimcore/blob/3ffaeac8d9f07f877ea4f469015be1774de447b9/tests/Model/Permissions/ModelAssetPermissionsTest.php#L488)and safe/consistent code, also seen on [search ](https://github.com/pimcore/pimcore/blob/3ffaeac8d9f07f877ea4f469015be1774de447b9/bundles/SimpleBackendSearchBundle/src/Controller/SearchController.php#L102)
https://github.com/pimcore/pimcore/blob/3ffaeac8d9f07f877ea4f469015be1774de447b9/bundles/SimpleBackendSearchBundle/src/Controller/SearchController.php#L403-L459

Initially tried to stick with the [tree](https://github.com/pimcore/admin-ui-classic-bundle/pull/767/commits/0ac73f6d49373f0c7b8062bd3c05b2cef9de9c07) "approach", but in the grid, there is no need to care about listing permission as we are not navigating within a tree

### DIFF
--- a/src/Controller/Admin/Asset/AssetController.php
+++ b/src/Controller/Admin/Asset/AssetController.php
@@ -1673,12 +1673,16 @@ class AssetController extends ElementControllerBase implements KernelControllerE
 
         if (!$this->getAdminUser()->isAdmin()) {
             $userIds = $this->getAdminUser()->getRoles();
-            $userIds[] = $this->getAdminUser()->getId();
-            $conditionFilters[] = ' (
-                                                    (select list from users_workspaces_asset where userId in (' . implode(',', $userIds) . ') and LOCATE(CONCAT(`path`, filename),cpath)=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
-                                                    OR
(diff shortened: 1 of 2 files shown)


mattamon (approved):
Tested it and got reliable results, also when testing the SQL Query directly in the database. LGTM

PR #773, opened 2024-11-28
Replace Request::get with explicit input sources

### LINKED ISSUE #771: Replace Request::get with explicit input sources
Related https://github.com/pimcore/pimcore/issues/14138

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/771


fashxp:
won't do that for classic ui bundle anymore.

PR #779, opened 2024-12-03
[Bug]: Fix show in tree

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/778


robertSt7:
@kingjia90 For me this doesn't fix the problem. Could you please take another look? TIA

kingjia90:
> @kingjia90 For me this doesn't fix the problem. Could you please take another look? TIA

Mhh, thank you for having a look, interesting...
Could it be a change in the core and a specific version combination of both installed is broken? I didn't see any regression/recent relevant change here in the admin ui to explain why it broke either, but locally it worked to me

kingjia90:
Closing in favor of https://github.com/pimcore/admin-ui-classic-bundle/pull/783

PR #745, opened 2024-11-13
Fix formatting for alias on relations

cancan101:
Closes #744

Issue #744, opened 2024-11-13
Operator Alias in Grid View on Many-to-One Relation Produces [object Object]


cancan101:
Actual:
![Image](https://github.com/user-attachments/assets/a81c69b9-838a-47ab-8b79-a67f4a7481fa)
![Image](https://github.com/user-attachments/assets/1d89f5df-9b40-4b45-bf28-c1475b5f5422)

Expected:
![Image](https://github.com/user-attachments/assets/9ba32a8a-0c92-47a2-8a8a-b1905848e807)
![Image](https://github.com/user-attachments/assets/40173b1b-c349-4ddb-9d95-f244e9f04a62)

Field:
![Image](https://github.com/user-attachments/assets/f9acaea9-9123-40d8-87c2-aba50ee51c36)

cancan101:
Whatever is calling Alias handles result from alias differently from result from DefaultValue. I need to track that down.

Basically without `Alias`, `getFullPath` (or similar) gets called and with alias it does not. I believe the rendering code path is very different for the case of an Alias vs direct field: https://github.com/pimcore/admin-ui-classic-bundle/blob/e71ee902ab1274a64d6b80d56af28f1855944dfd/src/Service/GridData/DataObject.php#L74-L220
something in the case of the direct field causes it to render as the full path.

fashxp:
Hi there, as we begin sunsetting admin-ui-classic in favor of Pimcore Studio, we won’t be able to fix or implement this issue. It should be improved in Studio though. 
Closing it out to keep things tidy – really appreciate your contribution and understanding!

PR #782, opened 2024-12-05
Revert "Fix formatting for alias on relations (#745)"

### LINKED ISSUE #780: Grid Operator | Relation Resolver broken
This is a regression of: https://github.com/pimcore/admin-ui-classic-bundle/pull/745/files (https://github.com/pimcore/admin-ui-classic-bundle/pull/745)

Using Operators like 
- Operator ObjectFIeld Getter
- Operator Any Getter

...won't work anymore and always return the `fullPath` of the given element.

Version: 1.6

cancan101:
This reverts commit 91d1903a946931d9899b4c81fccfd96061fdd46f.

Closes #780

The original fix was at the wrong location in the evaluation tree. Whatever is calling Alias handles result from alias differently from result from DefaultValue. I need to track that down.

Basically without alias, `getFullPath` (or similar) gets called and with alias it does not. Not sure why.

@fashxp 
@solverat
@la-lisa

### DIFF
--- a/src/DataObject/GridColumnConfig/Value/DefaultValue.php
+++ b/src/DataObject/GridColumnConfig/Value/DefaultValue.php
@@ -75,10 +75,6 @@ final class DefaultValue extends AbstractValue
             }
         }
 
-        if ($value instanceof ElementInterface) {
-            $value = $value->getFullPath();
-        }
-
         if (!$fieldDefinition instanceof Data) {
             return $this->getDefaultValue($value);
         }


solverat:
Yes, that fixes the issue. The original issue (#744) should be addressed at the operator "Alias" itself.

cancan101:
It might be more complicated than just fixing in the alias operator as the alias operator could itself be nested. I have to investigate that

cancan101:
I believe the rendering code path is very different for the case of an Alias vs direct field: https://github.com/pimcore/admin-ui-classic-bundle/blob/e71ee902ab1274a64d6b80d56af28f1855944dfd/src/Service/GridData/DataObject.php#L74-L220

something in the case of the direct field causes it to render as the full path.

robertSt7:
@cancan101 Thanks for reverting your PR. For further discussions I have re-opened the issue https://github.com/pimcore/admin-ui-classic-bundle/issues/744

Issue #784, opened 2024-12-06
[BUG]: Video Element not working


foxgreek:
Version: 1.6.3

When you try to edit a video you will get this error:
![Image](https://github.com/user-attachments/assets/a0cadec4-6e80-4b89-ad5c-987ee6f4e736)

You can reproduce this bug in the demo instance with the default video areabrick

blankse:
@foxgreek I created a PR: https://github.com/pimcore/admin-ui-classic-bundle/pull/785

robertSt7:
fixed by #785

PR #747, opened 2024-11-14
[Bugfix] Pass allowedTypes correctly on video editable

roland4432:
fix passing the correct allowedTypes from config to data (which is then passed to the helper to build the popup dialog)

PR #785, opened 2024-12-09
Fix: Video Element not working

blankse:
Fix: https://github.com/pimcore/admin-ui-classic-bundle/issues/784

### DIFF
--- a/public/js/pimcore/document/editables/video.js
+++ b/public/js/pimcore/document/editables/video.js
@@ -20,7 +20,7 @@ pimcore.document.editables.video = Class.create(pimcore.document.editable, {
     initialize: function($super, id, name, config, data, inherited) {
         $super(id, name, config, data, inherited);
 
-        data.allowedTypes = config.allowedTypes;
+        data.allowedTypes = config.allowedTypes ?? data.allowedTypes;
         this.data = data;
     },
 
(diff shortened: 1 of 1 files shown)


public/js/pimcore/document/editables/video.js
```diff
@@ -20,7 +20,6 @@ pimcore.document.editables.video = Class.create(pimcore.document.editable, {
     initialize: function($super, id, name, config, data, inherited) {
         $super(id, name, config, data, inherited);
 
-        data.allowedTypes = config.allowedTypes;
```
jdreesen, line 23: This was just added in #747. We should probably take another look at this ticket as well.

/cc @roland4432

blankse: @roland4432 @jdreesen The allowedTypes are already set from the config on the data in the editable php file. See https://github.com/pimcore/pimcore/blob/38728393ce02dfaa532f54c4fe99731b26638747/models/Document/Editable/Video.php#L1001

blankse: @roland4432 What was your `allowedTypes` configuration? Could it be that you used a wrong type in your array? Only an array with the values "asset", "youtube", "vimeo", and "dailymotion" is allowed. Otherwise, the configuration will be ignored. I think we should throw an exception when an incorrect type is used, rather than silently ignoring the configuration.

jdreesen: > I think we should throw an exception when an incorrect type is used, rather than silently ignoring the configuration.

:+1:

robertSt7:
@blankse Thanks for the fix

PR #786, opened 2024-12-09
[Task]: Followup #767 Grid permissions

kingjia90:
Without the CONCAT caused problems on permission of a specific element


kingjia90:
Given the table structure with clear distinction between `path` and `key`, the `CONCAT` could be avoided by a double condition
first to check the element itself, then the ones that have the same `path` as the fullpath for its children
image

PR #790, opened 2024-12-12
Allow calling shadowed method when using addMethods.

cancan101:
Currently this does not work:
```js
pimcore.object.search.addMethods({
    onRowContextmenu: function ($super, grid, record, tr, rowIndex, e, eOpts) {
        if (pimcore.currentuser.admin === true) {
            $super(grid, record, tr, rowIndex, e, eOpts);
            return;
        }
        ///
     }
});
```
Since `$super` looks for a method on the _superclass_ of `pimcore.object.search` and `onRowContextmenu` is defined on the class itself. With this PR, I can now do:
```js
pimcore.object.search.addMethods({
    onRowContextmenu: function ($this, grid, record, tr, rowIndex, e, eOpts) {
        if (pimcore.currentuser.admin === true) {
            $this(grid, record, tr, rowIndex, e, eOpts);
            return;
        }
        ///
     }
});
```

### DIFF
--- a/public/js/lib/class.js
+++ b/public/js/lib/class.js
@@ -64,9 +64,9 @@ var Class = (function() {
 
         for ( var i = 0, length = properties.length; i < length; i++) {
             var property = properties[i], value = source[property];
+            const method = value;
             if (ancestor && isFunction(value)
                     && argumentNames(value)[0] == "$super") {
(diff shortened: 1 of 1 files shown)


cancan101:
changing to draft for some more testing

PR #791, opened 2024-12-12
Field Type Table: Support drag & drop to change order of rows

BlackbitDevs:
This PR adds drag & drop support to change order of rows for field type `Table`.


mcop1:
Thank you!

PR #793, opened 2024-12-13
[BUG] fix permission check in GridHelperService for wrong query column if type is asset

### LINKED ISSUE #796: [Bug] list view for assets with non-admin user does not work
**Steps to reproduce:**

- Login with superuser
- Open an asset folder
- Switch the view to the list view

**Actual behaviour:**

- Error message appears: An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'key' in 'where clause'

**Expected behaviour:**

- Assets are shown in list view

**Environment:** 

- demo.pimcore.com

la-lisa:
Grid-proxy action fails if non admin user wants to access assets list in folder. Assets table has no column key only filename.

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -962,7 +962,7 @@ class GridHelperService
         return '(
             (`path` != "' . $path . '/" AND ' . $queryColumn .  ' != "' . $leaf . '")
             AND
-            `path` NOT LIKE "' . $fullpath . '%"
+            `path` NOT LIKE "' . $fullpath . '/%"
         )';
 
     }
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -936,7 +936,7 @@ class GridHelperService
         return '(
             (`path` = "' . $path . '/" AND ' . $queryColumn .  ' = "' . $leaf . '")
             OR
-            `path` LIKE "' . $fullpath . '/%"
+            `path` LIKE "' . $fullpath . '%"
         )';
     }
 
@@ -962,7 +962,7 @@ class GridHelperService
         return '(
             (`path` != "' . $path . '/" AND ' . $queryColumn .  ' != "' . $leaf . '")
             AND
-            `path` NOT LIKE "' . $fullpath . '/%"
+            `path` NOT LIKE "' . $fullpath . '%"
         )';
 
     }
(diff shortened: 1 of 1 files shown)


la-lisa:
have read the CLA Document and I hereby sign the CLA

la-lisa:

kingjia90:
Thank you for your PR!
Could you please rebase the branch to `1.7` as it should be a bugfix?

fashxp:
@la-lisa this PR has a lot unrelated commits in it. 
could you please rebase to 1.7 and fore-push, or create a new PR with your changes. I'd like to include that into todays release. 
thx very much.

la-lisa:
@fashxp done. Sorry did not check the history.

robertSt7:
@la-lisa It would be nice to have this in 1.6. Could you please re-base the branch to 1.6? Thanks a lot

src/Helper/GridHelperService.php
```diff
@@ -943,20 +944,25 @@ private function optimizedConcatLike(string $fullpath): string
...
         if ($onlyChildren) {
             return '`path` NOT LIKE "' . $fullpath . '/%"';
         }
 
         return '(
-            (`path` != "' . $path . '/" AND `key` != "' . $leaf . '")
+            (`path` != "' . $path . '/" AND ' . $queryColumn .  ' != "' . $leaf . '")
             AND
-            `path` NOT LIKE "' . $fullpath . '/%"
+            `path` NOT LIKE "' . $fullpath . '%"
```
fashxp: I don't think we need that change, right? 
It should not make any difference, but I would prefer to remove it and keep `"' . $fullpath . '/%"` as this prevents including unwanted folders. 

Example constellation would be: 
`foo/bar`
`foo/barmore` 

With old query, `foo/barmore` is not affected, with new query it is. 

As said, currently it does not make any difference, but it could bite us in future cases. 

I will do another PR for `optimizedConcatLike` to optimize that as well.

fashxp (changes requested):

src/Helper/GridHelperService.php
fashxp: reverted it now, so we can proceed. 
thx very much for your support!!

fashxp (approved):

robertSt7:
@la-lisa Thanks for the fix and your fast responses

PR #802, opened 2024-12-19
Update error modal to handle timedout.

cancan101:
Currently a timeout given this unhelpful message:
image

`timedout` is mentioned here: https://cdn.sencha.com/ext/commercial/4.2.2/docs/source/Connection.html


robertSt7:
@cancan101 Thanks for making the error message better

PR #803, opened 2024-12-19
Handle mime-type as system column.

cancan101:
Partially addresses: https://github.com/pimcore/pimcore/issues/17887
image


robertSt7:
@cancan101 Thanks for adding the mime-type to the grid

PR #804, opened 2024-12-26
Fix asset preview performance

cuca24:
Currently the performance of asset preview routes are still very poor with this fix: https://github.com/pimcore/admin-ui-classic-bundle/pull/767/files for non admin users. 

For example total number of assets: 500k it takes about ~13s for select query and additional ~13s for additional count query which barely avoids timeout of 30 seconds. 

@kingjia90 i added also call to grid helper functions for assets preview, which makes the performance the same for listing and preview. Is there any reason why we should not use it for preview like i did in PR?


cuca24:

kingjia90:
Thank you for the PR!

LGTM, the asset preview don't need to follow the list permission special rules (that is only applicable to tree view domain) and it should only display the permitted leafs.

As soon as the CLA is green, it could be merged.

cuca24:

PR #808, opened 2025-01-07
[Security]: Avoid user enumeration via reset password link

kingjia90:
Resolves https://github.com/pimcore/pimcore/security/advisories/GHSA-jjwj-6wr8-848h and https://github.com/pimcore/admin-ui-classic-bundle/security/advisories/GHSA-vr5f-php7-rg24


brusch (approved):
LGTM 👍

PR #809, opened 2025-01-08
Add concept of column readOnly to advancedManyToManyRelation

cancan101:
Introduce concept of column level readOnly to advancedManyToManyRelation.

Eventually this can be a config but for now this change allows users to subclass `pimcore.object.tags.advancedManyToManyRelation` and change the behavior.


mcop1:
Hello @cancan101 
since this is an improvement, could you please rebase the pr onto the current 2.x branch?

mcop1:
@cancan101 Never mind, changing the base branch was enough in this case. Thanks!

PR #810, opened 2025-01-15
Set php 8.3 as min

herbertroth:


composer.json
```diff
@@ -12,20 +12,20 @@
   "prefer-stable": true,
   "minimum-stability": "dev",
   "require": {
-    "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+    "php": "~8.3.0 || ~8.4.0",
```
alexz707, line 15: according our meeting in the morning this line should be removed completely right?

jdreesen: Please see: https://github.com/pimcore/generic-data-index-bundle/pull/276#issuecomment-2592722335

It is better to always explicitly define the dependencies of a package instead of relying on the definition of another package. This is especially true for the PHP version, as you can no longer change it once a package version has been released.

herbertroth: @alexz707 Yes, according to the meeting this should be removed. I figure out the that most of the  test depends on this(see workflows). So we can refactor the test or keep this. We will talk tomorrow in the meeting :-)

PR #815, opened 2025-01-24
Update endroid/qr-code to version 6

herbertroth:
- Update composer dependency to version 6
- Refactor QR code generation to use named constructor parameters
- Improve variable naming for better code readability


herbertroth:
Closes pimcore/product-management#735

PR #816, opened 2025-01-27
add nesbot/carbon to composer

alexz707:


src/DataObject/GridColumnConfig/Operator/DateFormatter.php
```diff
@@ -90,7 +90,7 @@ public function format(mixed $theValue): string
     {
         $timestamp = null;
         if (is_int($theValue)) {
-            $theValue = Carbon::createFromTimestamp($theValue);
+            $theValue = Carbon::createFromTimestamp($theValue, date_default_timezone_get());
```
jdreesen, line 93: Why is this necessary?

alexz707: Because the introduced a BC Break from V2 to V3:

Timezone for createFromTimestamp was previously date_default_timezone_get() in Carbon 2.

So to maintain compatibility this needs to be set.

PR #817, opened 2025-01-27
[Bug]: Fix potential leak when exporting

kingjia90:
Resolves https://pimcore.atlassian.net/browse/PEES-492, https://github.com/pimcore/service-operations/issues/87


blankse:
@kingjia90 I think we should put the fclose() to a finally block (https://www.php.net/manual/en/language.exceptions.php#language.exceptions.finally). So it will also be closed if a other exception is thrown and we have it only once.

kingjia90:
@blankse thank you, had the same thought but i wasn't sure, does it get affected by the return in the catch block?

blankse:
@kingjia90 See in the php docs:
`One notable interaction is between the finally block and a return statement. If a return statement is encountered inside either the try or the catch blocks, the finally block will still be executed.`

PR #819, opened 2025-01-29
Allow filtering Asset grid by id.

cancan101:
Allow filtering Asset grid by id. Makes the Asset grid consistent with the Object grid.


robertSt7:
@cancan101 Thanks for adding the filtering by id

PR #820, opened 2025-01-29
Allow filtering on Asset metadata.

cancan101:
This change is required to allow filtering of asset metadata,

see: https://github.com/pimcore/asset-metadata-class-definitions/pull/224

### DIFF
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -406,64 +406,79 @@ pimcore.element.helpers.gridColumnConfig = {
 
         const fieldInfo = this.grid.getColumns()[columnIndex].config;
 
+        let editor;
+        let tagType;
         if (this.objecttype === "object" || this.objecttype === "variant") {
             if (!fieldInfo.layout || !fieldInfo.layout.layout) {
                 return;
             }
 
-            const tagType = fieldInfo.layout.type;
-            const editor = new pimcore.object.tags[tagType](null, fieldInfo.layout.layout);
+            tagType = fieldInfo.layout.type;
+            editor = new pimcore.object.tags[tagType](null, fieldInfo.layout.layout);
             editor.setObject(this.object);
-            editor.updateContext({
-                containerType: "filterByRelationWindow"
-            });
-
-            const formPanel = Ext.create('Ext.form.Panel', {
-                xtype: "form",
-                border: false,
-                items: [editor.getLayoutEdit()],
-                bodyStyle: "padding: 10px;",
-                buttons: [
-                    {
-                        text: t("clear_relation_filter"),
-                        iconCls: "pimcore_icon_filter_condition pimcore_icon_overlay_delete",
-                        handler: function () {
-                            this.filterByRelationWindow.close();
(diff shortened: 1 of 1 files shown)


public/js/pimcore/element/helpers/gridColumnConfig.js
```diff
@@ -406,64 +406,79 @@ pimcore.element.helpers.gridColumnConfig = {
...
 
-            const tagType = fieldInfo.layout.type;
-            const editor = new pimcore.object.tags[tagType](null, fieldInfo.layout.layout);
+            tagType = fieldInfo.layout.type;
+            editor = new pimcore.object.tags[tagType](null, fieldInfo.layout.layout);
             editor.setObject(this.object);
-            editor.updateContext({
-                containerType: "filterByRelationWindow"
-            });
+        } else {
```
fashxp: I know it is different in other places, but should we do here an `else if (this.gridType === 'asset')` just to be sure?

fashxp (changes requested):

public/js/pimcore/element/helpers/gridColumnConfig.js
cancan101: should I have an else clause that raises some error, etc?

fashxp: would not do that, no - as it could be seen as a BC break (very unlikely someone uses that, but still...)

cancan101: I am not sure that would be any more breaking. if you do if / else approach and neither branch is hit, we will still get an exception on `editor.updateContext({` as editor won't have been set

fashxp: hm, but that wasn't the case before, right? 

maybe we need an else clause with a return statement?

cancan101: that issue does exist here: https://github.com/pimcore/admin-ui-classic-bundle/blob/3079471ccd83ec1598010cbe3b756891bb6c89da/public/js/pimcore/element/helpers/gridColumnConfig.js#L549-L564

the else clause assumes Asset but if you change that to an `else if` then there is a chance that no else is hit and the editor is not set.

cancan101: @fashxp I wanted to circle back on this remaining question.

fashxp: sorry, did't had the check the bigger context ... will do asap.

fashxp: so, now I had the chance to check: that is what I meant with 'I know it is different in other places' in my first comment. I know it would be broken on other places, but that doesn't mean we cannot make new parts more robust ;-)

I would do an else clause with a return statement - that should keep existing behavior for 'wrong' usages, right?

cancan101: the return statement in the else clause does look to be safe / an improvement. I will push

fashxp (approved):

fashxp:
thx very much!

PR #821, opened 2025-01-29
Fix: Add missing method “getComposerPackageName()” to get the correct bundle name and version

code-status-404:
This addition is necessary to query the correct composer bundle name and version programmatically.


src/PimcoreAdminBundle.php
```diff
@@ -23,13 +23,22 @@
...
 
 class PimcoreAdminBundle extends AbstractPimcoreBundle implements DependentBundleInterface
 {
+
+    use PackageVersionTrait;
+
+    public function getComposerPackageName(): string
+    {
+        return 'pimcore/admin-ui-classic-bundle';
+    }
```
jdreesen, line 40: Note that this method it not strictly necessary, because `PackageVersionTrait` determines this itself.

But explicitly defining it should be a bit quicker.

code-status-404: Thanks for the tip. Unfortunately, in this specific case it is necessary to define the method explicitly, as the name is not resolved in the trait because it is a “replaced” package. As a result, the bundle version cannot be retrieved programmatically.

jdreesen: Oh okay, good to know.

robertSt7:
@bk-hitcom Thanks a lot for the fix

PR #853, opened 2025-03-05
[Bug]: Missing Specific Settings when using FC in custom layouts 

kingjia90:
Resolves https://pimcore.atlassian.net/browse/PEES-527

~Not sure if it's correct, I can't find any occurence on why it got changed like that in the git history.~

### DIFF
--- a/public/js/pimcore/object/classes/data/fieldcollections.js
+++ b/public/js/pimcore/object/classes/data/fieldcollections.js
@@ -100,7 +100,7 @@ pimcore.object.classes.data.fieldcollections = Class.create(pimcore.object.class
             ]);            
         }
         this.specificPanel.add([
-            new Ext.ux.form.MultiSelect([{
+            {
                 xtype: "checkbox",
                 fieldLabel: t("lazy_loading"),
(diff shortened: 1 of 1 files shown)


fashxp:
hmm, at least the allowed types must not be extended (of changed at all) in custom layouts. that is also probably the reason, why all the options were deactivated.

kingjia90:
> hmm, at least the allowed types must not be extended (of changed at all) in custom layouts. that is also probably the reason, why all the options were deactivated.

ahh thank you, then it should just exclude that field, theoretically we could get the master layout value and let it pick from the "shortlist", it could be done by passing the layoutid and seem there is a parameter that can be passed
https://github.com/pimcore/admin-ui-classic-bundle/blob/74089384f8d9cb2a9ce02293de0c72e884545f00/src/Controller/Admin/DataObject/ClassController.php#L943
but as quickfix, this PR should be fine after https://github.com/pimcore/admin-ui-classic-bundle/pull/853/commits/373eac03d66daf45cc3b5db410021ce4581d4de0, they also "confirmed" in the ticket

PR #854, opened 2025-03-06
[Improvement] Add drag and drop to relations editable for documents

tobiashelmervalantic:
This pull request adds drag and drop to the relations editable for documents so they work the same as relations in objects.

image

Now you're able to drag and drop relations which is quite handy if you want to reorder a bigger list as using the arrows takes some time.


tobiashelmervalantic:

PR #862, opened 2025-03-11
Timezone: Consistent behavior for PHP and JavaScript

blankse:
Fix: https://github.com/pimcore/pimcore/issues/18142


fashxp:
thx very much

PR #871, opened 2025-03-18
Reverse relation fields: support column filtering

BlackbitDevs:
In analogy to #659 this PR adds column filtering also for reverse relation fields:
Bildschirmfoto 2025-03-18 um 08 03 10

Moreover column filtering did not work for fields which are configured to be `not editable`. With this PR this is possible (as filtering is no data change).


robertSt7:
@BlackbitDevs Thanks for adding the filtering

PR #875, opened 2025-03-21
Fix: Translate dashboard titles in toolbar and panel instead of displaying the “key”

code-status-404:
Currently, additional dashboards display the “key” as a visual name without the option to translate it. This option has been added.


fashxp:
thx very much.

PR #876, opened 2025-03-24
BugFix: Calculated Value Bool Type

torqdev:
When using the calculated value bool type it is using the input type instead of a checkbox.


fashxp (approved):

fashxp:
nice, thx very much!

PR #881, opened 2025-04-01
[Perspectives] Prevent objects from being copied to sections which do not allow objects of the given class

### LINKED ISSUE #865: Perspectives: Objects of wrong class can get added to wrong path via Copy/Paste or Cut/Paste
Steps to reproduce:
1. Open "PIM" perspective on https://demo.pimcore.com/admin/?perspective=PIM 
    It contains a custom view for `Car` objects on the left side and one for `Category` objects on the right side
    Image
2. Right-click category `/products` in the `Categories` custom view -> Cut
3. Right-click `/aston martin` on the `Cars` custom view and select "Paste > Paste cut-out element"

After pasting, the object is not visible in `/aston martin` in the `Cars` custom view. And it stays visible in the `Categories` custom view because this tree does not get refreshed.
After reloading Pimcore it is even worse: The object completely disappeared as it is not visible in the `Cars` custom view (because this custom view only shows objects of class `Car`) nor in the `Categories` custom view (because the object is in a different path now).

But when you switch to `Default` perspective, you will see that the object indeed got copied to `/aston martin` (or better `/Product Data/Cars/aston martin`):
Image

We need to prevent pasting elements to custom views which do not allow the class name of the copied element.

BlackbitDevs:
Resolves #865

### DIFF
--- a/public/js/pimcore/object/tree.js
+++ b/public/js/pimcore/object/tree.js
@@ -555,7 +555,7 @@ pimcore.registerNS("pimcore.object.tree");
                      var pasteMenu = [];
 
                      if (perspectiveCfg.inTreeContextMenu("object.paste")) {
-                         if (pimcore.cachedObjectId) {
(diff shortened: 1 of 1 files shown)


mcop1:
Thanks!

PR #886, opened 2025-04-05
fix: First unit not showing up in a dropdown

jantomicky:
This fixes the issue with the first quantity value unit not showing up in a dropdown select when attempting to set a base unit to another unit.

I have read the CLA Document and I hereby sign the CLA.


fashxp:
thx very much!

PR #887, opened 2025-04-07
Dispatching event within errorHandler function while executing upload asset via asset tree

BlackbitDevs:
We have a customer who would like to have a confirmation dialog at the upload event. I have added an event for this and implemented it in the error handler function that is called when the asset is uploaded via the asset tree.

The background is the intermediate duplicate check and throwing an exception if the asset (checksum identical) already exists in the system. The exception is filled with data accordingly and then evaluated in a JS plugin (customized).


mcop1:
Thanks!

PR #888, opened 2025-04-07
[Grid] Column filter for relational fields with width > 300 hides search icon

BlackbitDevs:
Steps to reproduce:

1. Create class with many-to-one relation field `Brand`
    Set `width` of field to `500`. 
2. Create object of this class
3. Open grid view of `/`, show field `Brand`
4. Use column filter for `Brand` -> Filter by relation

As the width of the popup window is only 500px, the input field is so wide that the looking glass icon does not fit anymore:
Bildschirmfoto 2025-04-07 um 16 35 18

If you set the field width to 600, the problem gets even worse: No buttons are shown at all:
Bildschirmfoto 2025-04-07 um 16 37 16

This PR ignores the set `width` in the relation filter dialogue but uses always 300px - the same value as if no `width` was configured for the field, see
https://github.com/pimcore/admin-ui-classic-bundle/blob/f08380863d34fb8aa8284ecb62d385c271408068/public/js/pimcore/object/tags/manyToOneRelation.js#L157-L161

Bildschirmfoto 2025-04-07 um 16 42 20

### DIFF
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -418,6 +418,8 @@ pimcore.element.helpers.gridColumnConfig = {
                 containerType: "filterByRelationWindow"
             });
 
+            editor.fieldConfig.width = 300;
+
             const formPanel = Ext.create('Ext.form.Panel', {
                 xtype: "form",
                 border: false,


mcop1:
Hello @BlackbitDevs ,

thanks for the pr. 
I was wondering if the method here https://github.com/pimcore/admin-ui-classic-bundle/blob/1.7/public/js/pimcore/element/helpers/gridColumnConfig.js#L452 was implemented for that reason as well. Maybe we could adapt that method for `manyToOneRelation`. What do you think?

astapc:
@BlackbitDevs Could you please respond within the next week?

PR #895, opened 2025-04-11
Fix Service::addTreeFilterJoins(): Argument #1 ($cv) must be of type …

### LINKED ISSUE #894: Pimcore\Model\Element\Service::addTreeFilterJoins(): Argument #1 ($cv) must be of type array, null given,
That error is raised here: https://github.com/pimcore/admin-ui-classic-bundle/blob/c57cb5ab365de192b342cddde9204613f7e9f117/src/Controller/Admin/DataObject/DataObjectController.php#L136

blankse:
…array, null given

Fix #894


fashxp:
thx

PR #905, opened 2025-04-14
[Improvement]: Remove delete button from element view toolbar

### LINKED ISSUE #904: [Improvement]: Remove the delete button from element view toolbar
### Improvement description
The suggestion is to remove the delete button and for a very simple reason: to avoid accidental, misunderstood, deletions.

Our user story/experience is this:
- user opens a data objects folder
- selects few data objects for deletion
- clicks on the delete button -> ERROR whole folder is deleted instead

That's exactly what happened to one of our client - mistakenly pressed on the toolbar's delete button. The correct action would've been to right click on the mouse and then select "Delete" from the opened context menu.

dolmit-tanel-paaro:
Resolves #904 

The user is expected to delete items only using the context menu that opens when you mouse right-click on an item in the sidebar element tree.

### DIFF
--- a/public/js/pimcore/asset/asset.js
+++ b/public/js/pimcore/asset/asset.js
@@ -230,9 +230,6 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
 
                 if (pimcore.helpers.checkIfNewHeadbarLayoutIsEnabled()) {
                     this.toolbarSubmenu.menu.add(deleteConfig);
-                } else {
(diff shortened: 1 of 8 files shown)


kingjia90:
Please see https://github.com/pimcore/admin-ui-classic-bundle/issues/904#issuecomment-3174017218 
If that comment is valid, i'd close this PR if you are ok with it

dolmit-tpaaro:
I'm okay closing this PR - #390 solves it nicely.

PR #908, opened 2025-04-16
[Documents] Areablock: Support controlsTrigger "click"

BlackbitDevs:
Currently the area block controls only support `controlsTrigger` values:
- `hover`: show area block controls when hovering the area brick with mouse cursor
- `fixed`: show all area brick controls

When you have a lot of area bricks, `hover` can be difficult to use as you could accidentically move the mouse over the wrong element. And `fixed` causes a lot of clutter with all the control bars being visible.

For this reason this PR adds `controlsTrigger: "click"`. This way the control bar gets shown if you click into an area brick.
For example we implemented absolutely positioned area bricks for a web-to-print application with the `click` approach:

https://github.com/user-attachments/assets/e9911ac0-9efd-4ee0-b989-4c0c8ddb67b6

### DIFF
--- a/public/js/pimcore/document/editables/areablock.js
+++ b/public/js/pimcore/document/editables/areablock.js
@@ -152,7 +152,7 @@ pimcore.document.editables.areablock = Class.create(pimcore.document.area_abstra
                             }
                         });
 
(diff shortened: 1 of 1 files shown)


robertSt7:
@BlackbitDevs Thanks for adding this option

PR #910, opened 2025-04-16
[Grid] Support filtering by multiple values in column filter

BlackbitDevs:
Imagine you are a data maintainer and get a list of SKUs provided. For all of them `fieldX` shall be set to `newValue`. This simple task is currently quite challenging since the SQL filter got removed in the grid.

This PR adds a way to filter a column by multiple values in the grid. You can provide a comma-separated list (CSV logic gets applied, so you can also escape commas if the filter value contains commas).

This way you can paste the comma-separated list of `SKU`s into the filter field with following syntax: 

`IN(a,b,c)` will search for `field='A' OR field='B' OR field='C'`

Bildschirmfoto 2025-04-16 um 16 39 22

It will find the matching objects and you can do the mass-editing.

As soon as you filter by multiple values, it now uses an exact search (not a `like` search). This is by design to not accidentally find elements which partly match. If you think it should still be a like search, we can change this of course.

Currently this PR only supports input and text fields but it could also be applied to numeric fields.


fashxp:
what if I want to search for a string with a `,` in it? this would not work then anymore, right?

BlackbitDevs:
> what if I want to search for a string with a , in it? this would not work then anymore, right?

As `str_getcsv()` is used for parsing you can quote values, e.g. `"value1,value2",value3` will search for `field = 'value1,value2' OR field='value3'`

cancan101:
From a UX standpoint, I don't like the magic of parsing w/ or w/o a comma differently. Further, the ability to escape using `"` is not immediately obvious to users (as is, they have to be trained on the use of `?` etc for prefix matches).

I do like the ability to filter with an `OR` as we have had similar needs. Perhaps there is a better way to do this with a UX change, e.g. if a comma is detected offer check mark control for `OR`?

image

BlackbitDevs:
> From a UX standpoint, I don't like the magic of parsing w/ or w/o a comma differently

You are right, this would speak for using the same filtering logic for single and multiple values - except that when providing multiple values, the single value filter logic gets applied multiple times. The only problem here is then when pasting a list of SKUs, that 1234 gets found when searching for 123 (because of the `LIKE` operator). The most common use-case for this multi-value filter is to get an exact list of products identified by copying the list from somewhere else (e.g. Excel). Finding false positives could result in data errors here (if after filtering and assuming that the found products are exactly the searched ones, a batch editing operation would be executed).

> Further, the ability to escape using " is not immediately obvious to users (as is, they have to be trained on the use of ? etc for prefix matches).

Yes, but this will be rarely necessary to know. Only if you are really searching for a comma, it would. And for this special case we could add this to documentation.

This extra checkbox is ultra-complicated to achieve with ExtJS.
WDYT of providing a filter value like `IN(a,b,c)`? This could be documented to search for exact matches.

cancan101:
>WDYT of providing a filter value like IN(a,b,c)? This could be documented to search for exact matches.

I like that a lot more.

BlackbitDevs:
Have changed this PR so that searching for `IN(a,b,c)` will search for `field='A' OR field='B' OR field='C'` while searching for `a,b,c` will continue to search for `field LIKE '%a,b,c%'` (same as current behaviour)

fashxp:
I like that approach much more as well

Can you please check the tests, thx

Jan-Walther-Basilicom:
@fashxp I have fixed the SonarQube issue about too much nesting.
But the Codeception problem seems to be unrelated to this PR:
> [InvalidArgumentException]                                                   
  `pimcore.encryption.secret` is not set.                                      
  Run `vendor/bin/generate-defuse-key` to generate a secret and set it as container parameter `pimcore.encryption.secret`.

fashxp:
mhm, we probably need to switch to pull_request_target for these actions. anyway.

fashxp:
looks good to me, thx very much!
was wondering if there would be some place to add docs for it (otherwise this will be quite a hidden feature). but I could not figure out a good place for it. 
so if somebody has an idea, a followup PR would be highly appreciated!

PR #916, opened 2025-04-25
[Bug] Grid filtering for ID does not work

### LINKED ISSUE #915: Grid filtering for ID does not work
**Steps to reproduce**

- Login as admin
- Open the grid for the cars
- Add a filter for an ID

**Actual behaviour**

- Error 500 appears

Image

**Expected behaviour**

- Filter should be applied without an error

**Environment**

- Platform Version
- Windows
- Google Chrome

mcop1:
Resolves #915 

Fixed other system field types as well.


herbertroth:
Pleas do not forget (now) there is a bugfix branch. So target must be 2.0 and an upstream merge is required.

PR #923, opened 2025-04-30
Fix invalid 2FA path name

saitho:
2FA login route name changed but was not adapted in template. Therefore, 2FA logins fail.

https://github.com/pimcore/admin-ui-classic-bundle/blob/10ea29130a41a9fefbcfc57b082f0ec5e3637a9b/src/Controller/Admin/LoginController.php#L398

### DIFF
--- a/templates/admin/login/two_factor_authentication.html.twig
+++ b/templates/admin/login/two_factor_authentication.html.twig
@@ -11,7 +11,7 @@
         </div>
     {% endif %}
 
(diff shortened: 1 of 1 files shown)


robertSt7:
As this is a bugfix, could you please execute `git rebase --onto origin/2.0 origin/2.x`?
Then you need to do a force-push:
`git push --force`
Thanks

saitho:
Ah, I see. Will do

robertSt7:
@saitho Thanks for fixing the path

PR #933, opened 2025-05-19
[PEES-556] Repeated batch edits of Images/Documents breaks EXT.js UI

robertSt7:
resolves https://github.com/pimcore/service-operations/issues/125


IronSean (approved):
Looks good in our testing. I wasn't sure if we should be doing more null checking and calling if possible but we didn't see any issues with this change when validating it.

PR #943, opened 2025-05-28
(Bugfix) Creation of dynamic property is deprecated

cgonzalezbermudez:
as pimcore (platform-version v2024.4) is compatible with php 8.3 and 8.4, an issue is appearing with the message 
Deprecated: Creation of dynamic property Pimcore\Model\User\Workspace\Asset::$path is deprecated in /var/www/html/vendor/pimcore/admin-ui-classic-bundle/src/Controller/Admin/UserController.php on line 713
and blocking the capacity to modify roles with has paths definitions in the workspaces

### DIFF
--- a/public/js/pimcore/settings/user/workspace/asset.js
+++ b/public/js/pimcore/settings/user/workspace/asset.js
@@ -32,10 +32,10 @@ pimcore.settings.user.workspace.asset = Class.create({
 
         var availableRights = ["list","view","publish","delete","rename","create","settings","versions","properties"];
         var gridPlugins = [];
-        var storeFields = ["path"];
+        var storeFields = ["cpath"];
 
         var typesColumns = [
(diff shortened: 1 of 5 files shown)


cgonzalezbermudez:

kingjia90:
recheck

fashxp:
@carlosgonzalezpdn it seems that in that PR are now commits which don't belong to that PR? 
could you please double check and have a look. Thx

kingjia90:
Closing due to inactivity, branch basing and CLA issues

kingjia90:
In the meanwhile, i've provided an alternative on https://github.com/pimcore/admin-ui-classic-bundle/pull/986 that avoids the refactor of `path` to `cpath` in js.
`$workspace->path` is not used right after being created, so should be just fine

PR #946, opened 2025-06-05
Bugfix: Allow defaultvalue `0` for numeric fields

turbo-ele:
Currently, the default value `0` is being ignored.


fashxp:
thx very much!

PR #947, opened 2025-06-05
Marking classification fields as dirty when changed in the grid-view

turbo-ele:
When changing a value in the classificationstore grid, it was not being marked as dirty

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectActionsTrait.php
+++ b/src/Controller/Admin/DataObject/DataObjectActionsTrait.php
@@ -209,6 +209,8 @@ trait DataObjectActionsTrait
                             }
                         }
 
+                        $object->markFieldDirty($field);
+
                         $activeGroups = $classificationStoreData->getActiveGroups() ?: [];
                         $activeGroups[$groupId] = true;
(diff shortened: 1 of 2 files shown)


fashxp:
what is the consequence of that? where does the problems appear...

fashxp:
:ping_pong: @turbo-ele

valantic-cx-alps:
In our case, we wanted to be able to detect in an event-subscriber, if a value in the classificationstore was changed. This works, when the value is changed in the edit ui of the data-object itself, but not, when a value is changed directly in the datgrid.
Therefore, the classificationstore field should be marked as dirty, for an object, when a value is changed in the grid as well.

turbo-ele:
@fashxp 
In our case, we wanted to be able to detect in an event-subscriber, if a value in the classificationstore was changed. This works, when the value is changed in the edit ui of the data-object itself, but not, when a value is changed directly in the datgrid.
Therefore, the classificationstore field should be marked as dirty, for an object, when a value is changed in the grid as well.

fashxp:
thx very much for the clarification.

PR #949, opened 2025-06-06
Support having the same language with multiple locales for preview

turbo-ele:
Otherwise it is not possible to differ between locales of the same language.
E.g. having de_DE, de_CH, de_AT will always lead to just having 'German' as select option

### DIFF
--- a/src/Service/PreviewGenerator.php
+++ b/src/Service/PreviewGenerator.php
@@ -82,7 +82,7 @@ class PreviewGenerator implements PreviewGeneratorInterface
 
         $locales = [];
         foreach (Tool::getValidLanguages() as $locale) {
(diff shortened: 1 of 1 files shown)


fashxp:
hmm, for me this fixes the language selection, but the preview loading it self seems not considering the region part of the locale?!

turbo-ele:
Selecting the full locale e.g. `de_AT` is triggering the DataObjectController::previewAction with a query parameter `locale=de_AT` which is then being passed to the `params` array into the LinkGenerator.

Of course, it then needs to be processed the right way.

fashxp:
and in our demo that is not the case ... thx for clarification and the fix!

PR #956, opened 2025-06-18
[Workflow Transitions] Apply default value for additional fields

BlackbitDevs:
Currently default values for additional fields in workflows do not get applied. This PR changes this by calling `applyDefaultValue()`, similar to https://github.com/pimcore/admin-ui-classic-bundle/blob/6ad887e762b2c3b5b02527b0d61080394436d2b8/public/js/pimcore/object/helpers/edit.js#L340-L342

### DIFF
--- a/public/js/pimcore/workflow/transitionPanel.js
+++ b/public/js/pimcore/workflow/transitionPanel.js
@@ -272,8 +272,10 @@ pimcore.workflow.transitionPanel = Class.create({
                         field.setHeight(100);
                     }
 
-                    if (typeof field.applyDefaultValue !== "undefined") {
-                        field.applyDefaultValue();
+                    if (typeof field.setValue !== "undefined") {
+                        field.setValue(c.defaultValue);
+                    } else {
+                        field.config.items[0]?.setValue(c.defaultValue);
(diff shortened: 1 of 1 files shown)


public/js/pimcore/workflow/transitionPanel.js
```diff
@@ -272,6 +272,9 @@ pimcore.workflow.transitionPanel = Class.create({
                         field.setHeight(100);
                     }
 
+                    if (typeof field.applyDefaultValue !== "undefined") {
+                        field.applyDefaultValue();
```
robertSt7: @BlackbitDevs `field.applyDefaultValue` is always `undefined`, the `applyDefaultValue` function is in the `tag`, but when we call it there the default value isn't displayed in the workflow execution. Could you please take a look at this? Thx

astapc:
@BlackbitDevs  Could you please review and respond to [this pull request](https://github.com/pimcore/admin-ui-classic-bundle/pull/956#pullrequestreview-2978158526) within the next week?

robertSt7:
@BlackbitDevs Thanks for the PR

PR #957, opened 2025-06-18
BugFix: Classification Store Attribute Not Displaying Correct Value In Grid

torqdev:
When inheritance is enabled, the classification store attribute value is not being displayed correctly in the grid. 

Check to see if classification store attribute is an array before trying to access the `value` key. If not array, check to see if attribute has a value (similar to the logic when getting the inherited data: https://github.com/pimcore/admin-ui-classic-bundle/blob/1.7/src/Service/GridData/DataObject.php#L352)

### DIFF
--- a/src/Service/GridData/DataObject.php
+++ b/src/Service/GridData/DataObject.php
@@ -180,7 +180,7 @@ class DataObject extends Element
                     }
 
                     // because the key for the classification store has not a direct getter, you have to check separately if the data is inheritable
-                    if (str_starts_with($key, '~') && empty($data[$key]['value'])) {
+                    if (str_starts_with($key, '~')) {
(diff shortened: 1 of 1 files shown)


robertSt7:
@torqdev I am not able to reproduce this. Could you please provide some more info like the type in the classification store you used? Thanks

torqdev:
Hi @robertSt7 - If inheritance is turned on, and the parent object has a value it does show on child objects as expected. But once the child value is overridden it only shows the inherited value in the grid.

Added classification store attributes to grid, and added values to parent attributes. Values are inherited to child accessory part. Editing the child accessory part continues to only display the parent inherited value.
image

Opening up child accessory part shows correct new values after editing in grid.
image

robertSt7:
@torqdev Thanks for this fix

PR #965, opened 2025-07-09
Allow filtering on reverse relation

cancan101:
Reverts #375

Depends on: https://github.com/pimcore/pimcore/pull/18036

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -335,7 +335,7 @@ class GridHelperService
                         if (is_array($filter['value'] ?? false)) {
                             $fieldConditions = [];
                             foreach ($filter['value'] as $filterValue) {
-                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, []);
+                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
                             }
 
                             if (!empty($fieldConditions)) {
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -335,14 +335,14 @@ class GridHelperService
                         if (is_array($filter['value'] ?? false)) {
                             $fieldConditions = [];
                             foreach ($filter['value'] as $filterValue) {
-                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
+                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, []);
                             }
 
                             if (!empty($fieldConditions)) {
                                 $conditionPartsFilters[] = '(' . implode(' OR ', $fieldConditions) . ')';
                             }
                         } else {
-                            $conditionPartsFilters[] = $field->getFilterCondition($filter['value'] ?? null, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
+                            $conditionPartsFilters[] = $field->getFilterCondition($filter['value'] ?? null, $operator, ['tablePrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
                         }
                     } elseif (in_array($filterField, $systemFields)) {
                         // system fields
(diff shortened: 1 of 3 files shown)


src/Helper/GridHelperService.php
```diff
@@ -335,14 +335,14 @@ public function getFilterCondition(string $filterJson, ClassDefinition $class, ?
                         if (is_array($filter['value'] ?? false)) {
                             $fieldConditions = [];
                             foreach ($filter['value'] as $filterValue) {
```
cancan101: FU to https://github.com/pimcore/admin-ui-classic-bundle/commit/8d462971862ea791988979fe9f2587c5ddfc14da, I honestly don’t understand why `brickPrefix` has to be passed anywhere outside this [if block](https://github.com/pimcore/admin-ui-classic-bundle/blob/7bf40f83bc1e93bd58decc2db98a85015a5d8461/src/Helper/GridHelperService.php#L294).

fashxp: Me neither. They were added here: https://github.com/pimcore/pimcore/pull/8829
So I would prefer to not to remove the `brickPrefix`

src/Helper/GridHelperService.php
```diff
@@ -335,14 +335,14 @@ public function getFilterCondition(string $filterJson, ClassDefinition $class, ?
...
-                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
+                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, []);
```
fashxp:
```suggestion
                                $fieldConditions[] = $field->getFilterCondition($filterValue, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
```

src/Helper/GridHelperService.php
```diff
@@ -335,14 +335,14 @@ public function getFilterCondition(string $filterJson, ClassDefinition $class, ?
...
                             }
 
                             if (!empty($fieldConditions)) {
                                 $conditionPartsFilters[] = '(' . implode(' OR ', $fieldConditions) . ')';
                             }
                         } else {
-                            $conditionPartsFilters[] = $field->getFilterCondition($filter['value'] ?? null, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
+                            $conditionPartsFilters[] = $field->getFilterCondition($filter['value'] ?? null, $operator, ['tablePrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
```
fashxp:
```suggestion
                            $conditionPartsFilters[] = $field->getFilterCondition($filter['value'] ?? null, $operator, ['brickPrefix' => ($tablePrefix ? $tablePrefix . '.' : null), 'tablePrefix' => ($tablePrefix ? $tablePrefix . '.' : null)]);
```

fashxp (changes requested):

fashxp:
@cancan101 please also bump the pimcore/pimcore dependency to 12.2

src/Helper/GridHelperService.php
cancan101: `brickPrefix` isn't right outside of the block as that isn't a prefix for a brick.

src/Helper/GridHelperService.php
cancan101: I am depending on `tablePrefix` here: https://github.com/pimcore/pimcore/pull/18036/files#diff-2b30de942edd58bfcbda8ca04a54619d65bcb861da86cfe7cd22699570bee8d3R219

fashxp: yes I know ... my suggestion would additionally add the needed `tablePrefix`, but  not replace/remove the `brickPrefix`. this should work right? at least in my tests it did.

cancan101: ah, i see, yes that would work. Are there an concerns with passing a value in `brickPrefix` when that isn't actually the brick prefix?

fashxp: At least there is no known problem related to this. 
Normally I would not say that and would investigate: But since admin-ui-classic is about to be sun-set, I would like to avoid introducing new bugs with refactorings anymore.

fashxp (approved):
:heart: looks good to me now, thx very much!

PR #969, opened 2025-07-25
fix(json-translations-system): used array_replace instead of array_merge which results wrong key => value pairs for numeric keys

patrickbitzer:
# Issue

In Pimcore it is possible to translate elements in the Pimcore Admin Interface using Admin Translation. For example, if a folder is to be named differently than in the original.
 
Now we have discovered the following phenomenon and unfortunately I cannot determine the cause. The fact is, however, that it is a translation problem.
 
We use CoreShop and there, as with the Ecommerce Framework, a folder structure is created for the orders or shopping carts --> year/month/day
 
Lev had tested the order process and noticed the following when checking the admin interface. A folder with the key “200” was displayed within the folder structure, both under 2025/06/ and under 2025/07. After closer examination, it was the folders with the key “11”. The path is correct if you look at the folder information.
 
Strangely enough, after emptying the Pimcore cache (bin/console pi:ca:cl), the designation “11” could be seen again, but now other numerical designations such as ‘10’ were displayed as “200”.

image

image

# Solution
By using array_replace instead of array_merge, I could fix the problem.

### DIFF
--- a/src/Controller/Admin/MiscController.php
+++ b/src/Controller/Admin/MiscController.php
@@ -93,7 +93,7 @@ class MiscController extends AdminAbstractController
         }
 
         foreach (['admin', 'admin_ext'] as $domain) {
-            $translations = array_merge($translations, $translator->getCatalogue($language)->all($domain));
+            $translations = array_replace($translations, $translator->getCatalogue($language)->all($domain));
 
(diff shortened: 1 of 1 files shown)


astapc:
Closed due https://github.com/pimcore/admin-ui-classic-bundle/pull/968 - will be upmerged from there

PR #977, opened 2025-08-07
[Bug, EC] PEES-674: Bug: Classification Store Filtering In Grid

kingjia90:
Resolves https://github.com/pimcore/service-operations/issues/200

Requires https://github.com/pimcore/pimcore/pull/18612

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -118,11 +118,19 @@ class GridHelperService
                     $field = \Pimcore\Model\DataObject\Classificationstore\Service::getFieldDefinitionFromJson($definition, $type);
 
                     if ($field instanceof Model\DataObject\ClassDefinition\Data) {
+                        $featureJoin = [
+                            'fieldname' => $fieldName,
+                            'groupId' => $groupId,
+                            'keyId' => $keyid,
+                            'language' => $language
+                        ];
+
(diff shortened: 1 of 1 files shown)


src/Helper/GridHelperService.php
```diff
@@ -454,8 +467,13 @@ public function addGridFeatureJoins(DataObject\Listing\Concrete $list, array $fe
                     }
                     $alreadyJoined[$mappedKey] = 1;
 
+                    if (isset($featureJoin['value2'])){
+                        $secondValue = ' and ' . $mappedKey . '.value2 = ' . $db->quote($featureJoin['value2']);
```
kingjia90: need to introduce this as join condition for the measuring unit
image

kingjia90: The problem is that the query is like having an ALIAS ambiguity
image
Can't use the `cskey_technicalAttributes_5_8.value2` in the `HAVING` because of the select `cskey_technicalAttributes_5_8.value AS cskey_technicalAttributes_5_8`

kingjia90: image
and luckily the unit is checked once in the filter, even if you use both greater than and less than.

kingjia90:
Cherry picked in 1.7 on https://github.com/pimcore/admin-ui-classic-bundle/commit/68c0e77e5164f6349e704a9fbbbc0216876e456a

PR #990, opened 2025-09-05
[Bug] Fix strlen(): Argument #1 ($string) must be of type string, array given in SubString Operator

### LINKED ISSUE #919: [Bug]  strlen(): Argument #1 ($string) must be of type string, array given
```
TypeError: strlen(): Argument #1 ($string) must be of type string, array given
#29 /vendor/pimcore/admin-ui-classic-bundle/src/DataObject/GridColumnConfig/Operator/Substring.php(70): strlen
#28 /vendor/pimcore/admin-ui-classic-bundle/src/DataObject/GridColumnConfig/Operator/Substring.php(70): Pimcore\Bundle\AdminBundle\DataObject\GridColumnConfig\Operator\Substring::getLabeledValue
#27 /vendor/pimcore/pimcore/models/DataObject/Service.php(539): Pimcore\Model\DataObject\Service::Pimcore\Model\DataObject\{closure}
#26 /vendor/pimcore/pimcore/models/DataObject/Service.php(1969): Pimcore\Model\DataObject\Service::useInheritedValues
```

Image

Not entirely sure how I triggered this but assume it involves using substring operator on an array of arrays.

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/919

Enhancing "tolerance/gracefulness" for unsupported combinations, eg. trying to implode in case is an array of strings or convert objects with `__toString()` (usually would be returning the path) to a string.
Changing to more UTF8-safe multi-byte character functions

### DIFF
--- a/src/DataObject/GridColumnConfig/Operator/Substring.php
+++ b/src/DataObject/GridColumnConfig/Operator/Substring.php
@@ -69,9 +69,25 @@ final class Substring extends AbstractOperator
                     if (!$childValue) {
                         continue;
                     }
-                    // try to implode if it's a plain array of strings
-                    if (is_array($childValue)){
-                        $childValue = implode(' ', array_map('strval', array_values($childValue)));
+
+                    if (!is_string($childValue)) {
+                        if (is_array($childValue)) {
+                            $childValue = implode(
+                                ' ',
+                                array_map(
+                                    fn($v) => is_scalar($v) ? (string)$v : '',
+                                    array_values($childValue)
+                                )
+                            );
+                        } elseif (is_object($childValue)) {
+                            // Convert object safely, prefer __toString if available
+                            $childValue = method_exists($childValue, '__toString')
+                                ? (string) $childValue
+                                : '';
+                        } else {
+                            // fallback for other types (int, float, bool, null)
+                            $childValue = (string) $childValue;
+                        }
                     }
 
                     $showEllipses = $useEllipses && mb_strlen($childValue) > ($start + $length);
@@ -91,7 +107,7 @@ final class Substring extends AbstractOperator
             if ($isArrayType) {
                 $result->value = $valueArray;
             } else {
-                $result->value = $valueArray[0];
+                $result->value = $valueArray[0] ?? '';
             }
         }
 
--- a/src/DataObject/GridColumnConfig/Operator/Substring.php
+++ b/src/DataObject/GridColumnConfig/Operator/Substring.php
@@ -80,12 +80,11 @@ final class Substring extends AbstractOperator
                                 )
                             );
                         } elseif (is_object($childValue)) {
-                            // Convert object safely, prefer __toString if available
                             $childValue = method_exists($childValue, '__toString')
                                 ? (string) $childValue
                                 : '';
                         } else {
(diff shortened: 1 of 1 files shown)


src/DataObject/GridColumnConfig/Operator/Substring.php
```diff
@@ -84,13 +91,35 @@ public function getLabeledValue(array|ElementInterface $element): ResultContaine
...
     }
 
+    private function convertToString(mixed $value): string
+    {
+        if (is_array($value)) {
+            $output = implode(
+                ' ',
+                array_map(
+                    fn($v) => is_scalar($v) ? (string)$v : '',
+                    array_values($value)
```
jdreesen: This is unnecessary, as `implode` does not care about array keys.

```suggestion
                    $value
```

src/DataObject/GridColumnConfig/Operator/Substring.php
```diff
@@ -84,13 +91,35 @@ public function getLabeledValue(array|ElementInterface $element): ResultContaine
...
         return $result;
     }
 
+    private function convertToString(mixed $value): string
+    {
+        if (is_array($value)) {
+            $output = implode(
+                ' ',
+                array_map(
+                    fn($v) => is_scalar($v) ? (string)$v : '',
```
jdreesen: Objects with `__toString` as array values can never happen?

If it is possible:
```suggestion
                    fn($v) => is_scalar($v) || $v instanceof \Stringable ? (string)$v : '',
```

or maybe better recursively call this method on all array values?

src/DataObject/GridColumnConfig/Operator/Substring.php
```diff
@@ -84,13 +91,35 @@ public function getLabeledValue(array|ElementInterface $element): ResultContaine
...
+                    array_values($value)
+                )
+            );
+        } elseif (is_object($value)) {
+            $output = method_exists($value, '__toString')
```
jdreesen:
```suggestion
            $output = $value instanceof \Stringable
```

(every object that implements this method automatically also implements this interface, even if not explicitly declared)

src/DataObject/GridColumnConfig/Operator/Substring.php
kingjia90: Thank you for all the feedback, it also crossed my mind of some recursion, but it gets into some area where the data is really out of the scope or usefulness for a substring functionality, also i was not sure if a space separator for implode would be good.

src/DataObject/GridColumnConfig/Operator/Substring.php
kingjia90: ohhh.. i didn't know that or i just forgot about this magic , thank you for the insight
it's exactly what i needed, as `method_exists` is not really optimal other than for BC/FC and compatibility stuff 😄

jdreesen:
You don't need to actually change anything to trigger the tests. Adding an empty commit (`git commit --allow-empty -m "Trigger Pipeline"`) should suffice.

kingjia90:
Thank you 😊 something new learnt. I've just wanted something quick to do it on this page GUI, but i realize i just could re-run (which i did in another one). For older PRs, it won't let re-run.

Issue #1002, opened 2025-09-17
[Bug]: Invalid translations for numeric translation keys


nbb-content:
Description see:  https://github.com/pimcore/pimcore/issues/18171
Tagged with: PR Welcome

Since we faced the same problem and were able to prove that the proposed solution works perfectly, we would like to provide the PR for it.

nbb-content:
PR provided.

nbb-content:
Fyi, it seems it was already solved with https://github.com/pimcore/admin-ui-classic-bundle/pull/968 😅
I guess because it was reported initially in the main Pimcore project it was not detected as duplicate. 🙂

fashxp:
fixed via https://github.com/pimcore/admin-ui-classic-bundle/pull/968

PR #1003, opened 2025-09-17
Use array_replace to keep numeric and non-numeric array keys

nbb-content:
This is a fix for the issue described here: https://github.com/pimcore/admin-ui-classic-bundle/issues/1002


fashxp:
closed in favor of https://github.com/pimcore/admin-ui-classic-bundle/pull/968 

thx for you contribution anyway!!

Issue #655, opened 2024-08-08
[Bug]: TinyMCE  - Shared Translations - surrounding <p> Tag


ctippler:
Pimcore Version: 11.2.0

Steps to reproduce:
The problem is, that TinyMCE adds a surrounding `<p></p>` Tag on translations where we don't want/need the` <p> `Tag. 

1) Go to "Tools" > "Translations" > "Translations"
2) Add an Translation with an HTML Code in it "E.g" We want the following code as Translation
`This is <a href="{link}">a sample</a> sentence.`
3) save/update the translation
4) Re edit the translation (now you are forced to use the TinyMCE Wysiwyg Editor) and change the text. Then update/save the translation. Now TinyMCE adds a surrounding "<p>..</p>" tag which we don't want. 

As TinyMCE remove the option "forced_root_block" (https://github.com/tinymce/tinymce/discussions/7342) it seems there  is no way around this issue. 

So a solution would be to let the input field editable and display both buttons "Edit as HTML" and "Edit as Text". So the user can decide which Editor he wants.

Threfore we would just need to remove this part
https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/public/js/pimcore/settings/translation.js#L622-L636

What do you think? Is this ok/ a reasonable way for you.

Regards,
Christian

fashxp:
Hmm, I can't think of a better solution either? Anyone else? 
Or could you provide a PR for that?

kingjia90:
Thank you for reporting and suggesting the fix.

I've tried but couldn't also find a viable solution , so applying the suggested modification by https://github.com/pimcore/admin-ui-classic-bundle/pull/1009

Since it's for sunsetting, there would be no room for introducing alternative editors that work well with simple and more "inline" (without paragraphs,div,etc..) html (dunno, even adding a Markdown compatible alternative), i would see it as a viable compromise to let users manually edit in text mode.

kingjia90:
Resolved by https://github.com/pimcore/admin-ui-classic-bundle/pull/1009

PR #1008, opened 2025-09-22
[Bug]: Clean up <p> WYSIWYG editor output from Translation 

kingjia90:
Resolves https://github.com/pimcore/admin-ui-classic-bundle/issues/655

### DIFF
--- a/public/js/pimcore/settings/translationEditor.js
+++ b/public/js/pimcore/settings/translationEditor.js
@@ -152,9 +152,7 @@ pimcore.settings.translation.editor = Class.create({
 
         document.addEventListener(pimcore.events.changeWysiwyg, function (e) {
             if (this.editableDivId === e.detail.e.target.id) {
-                this.value = e.detail.data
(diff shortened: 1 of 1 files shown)


kingjia90:
Draft: not good, it would be wrong if there are more than 1 <p> element

public/js/pimcore/settings/translationEditor.js
```diff
@@ -152,7 +152,7 @@ pimcore.settings.translation.editor = Class.create({
 
         document.addEventListener(pimcore.events.changeWysiwyg, function (e) {
             if (this.editableDivId === e.detail.e.target.id) {
-                this.value = e.detail.data;
+                this.value = e.detail.data.replace(/<p>/g, "").replace(/<\/p>/g, "</ br>");
```
jdreesen: There is no `</ br>` tag. Maybe you meant `<br />`? But this is also no HTML5, which just uses `<br>`.

kingjia90: ah yes, thank you, it was a typo, but didn't work as well 😞  closing the PR in favor of https://github.com/pimcore/admin-ui-classic-bundle/pull/1009

PR #1014, opened 2025-09-30
Use Attribute instead of Annotation to define route. Add pimcore_admin_element_islocked

muratbinerbay:
Pimcore 12.2.1 with admin-ui-classic-bundle 2.2.0 fails to check if an element is locked because the route `pimcore_admin_element_islocked` was never defined in ElementController:

image

As Symfony Annotation is deprecated the route definition for `lockElementAction` does not work

Related to https://github.com/pimcore/admin-ui-classic-bundle/pull/724

### DIFF
--- a/public/js/pimcore/helpers.js
+++ b/public/js/pimcore/helpers.js
@@ -766,8 +766,7 @@ pimcore.helpers.lockManager = function (cid, ctype, csubtype, data) {
                     }
                 });
             } else {
-                pimcore.helpers.closeElement(lock[0], lock[1]);
-                pimcore.helpers.openElement(lock[0], lock[1], lock[2]);
+                pimcore.layout.refresh();
             }
         }.bind(this, arguments));
 };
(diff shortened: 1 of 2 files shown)


muratbinerbay:
Had to call `activeTab.initialConfig.document.reload() / activeTab.initialConfig.object.reload()` because the consecutive calls for close & open was leading to this error:

Screenshot 2025-09-30 at 1 45 06 pm

public/js/pimcore/helpers.js
```diff
@@ -766,8 +768,7 @@ pimcore.helpers.lockManager = function (cid, ctype, csubtype, data) {
                     }
                 });
             } else {
-                pimcore.helpers.closeElement(lock[0], lock[1]);
-                pimcore.helpers.openElement(lock[0], lock[1], lock[2]);
+                pimcore.helpers.refreshElement();
```
kingjia90, line 777: not related to this PR, but something doesn't make sense from #724, when replying "No" to the popup that warns about element being locked, it would reload and unlock

kingjia90 (commented):
Thank you for the PR! 

Indeed it wasn't working properly

kingjia90:
Cherry picked in `2.2` by https://github.com/pimcore/admin-ui-classic-bundle/commit/162ad4713e39970900d26e563305b2ccacc349f3

PR #1020, opened 2025-10-03
Bugfix: Asset Numeric IN Filtering

torqdev:
When filtering assets, specifically numeric fields like `id`, the values were not being applied with the IN filter in the query. This makes it so users could not filter by multiple asset ids like they could with data objects.

When filtering via numeric values, it now checks if its a `in` filter operator, then attempts to check to see if the value is already an array. If not, it uses similar logic to data object numeric parsing to split on all non-numeric values to get an array. Also added a cast on the `db->quote` value to make sure all values are strings.

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -840,6 +840,16 @@ class GridHelperService
                         $operator = '=';
                     } elseif ($filterOperator == 'in') {
                         $operator = 'IN';
+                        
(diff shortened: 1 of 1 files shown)


robertSt7:
@torqdev Thanks for fixing the filtering

kingjia90:
cherry picked in `1.7` by https://github.com/pimcore/admin-ui-classic-bundle/commit/16437c6ec44e61bceac7ab0d791fde399c95778f

PR #1022, opened 2025-10-06
Asset folder preview: do not render preview images on-the-fly if they…

brusch:
… do not exist yet - as this could potentially break the entire hosting infrastructure when dealing with huge image files (>500MB)

This PR shows a placeholder instead: 

image


bluvulture:
@brusch we got a question would we also be able to apply it to 1.7.x versions?

brusch:
Yeap, just rebased / changed target on 1.7 branch and changed the milestone to 1.7.13.

robertSt7 (approved):

PR #1023, opened 2025-10-06
Fix: Filtering for checkbox in assets

robertSt7:
resolves https://pimcore.atlassian.net/browse/PEES-792

### DIFF
--- a/src/Helper/GridHelperService.php
+++ b/src/Helper/GridHelperService.php
@@ -777,6 +777,7 @@ class GridHelperService
             $filters = json_decode($filterJson, true);
             foreach ($filters as $filter) {
                 $operator = '=';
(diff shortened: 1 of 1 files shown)


src/Helper/GridHelperService.php
```diff
@@ -856,7 +857,8 @@ public function prepareAssetListingForGrid(array $allParams, User $adminUser): M
                     $operator = 'IN';
                 } elseif ($filterType == 'boolean') {
                     $operator = '=';
-                    $filter['value'] = (int) $filter['value'];
+                    $notSubselect = 'NOT';
```
```suggestion
                    if ((int) $filter['value'] === 0) {
                        $notSubselect = 'NOT';
                    }
```

PR #1024, opened 2025-10-06
Bugfix: Update regex, and add new translation with regex params.

torqdev:
The object brick regex does not match the regex that is used in the backend. The object brick checks to see if its a forbidden name(https://github.com/pimcore/pimcore/blob/12.2/models/DataObject/Objectbrick/Definition.php#L514), which the object brick class extends field collection which points to: https://github.com/pimcore/pimcore/blob/12.x/models/DataObject/Fieldcollection/Definition.php#L262-L273.

As well, added a more detailed version of the translation which specifies the requirements when adding a new brick.


fashxp:
actually a followup to https://github.com/pimcore/pimcore/pull/17881

thx very much!

PR #1025, opened 2025-10-07
BugFix: Fix openTabs limiting logic to always keep only the last 10 items

hopleus:
Previously, the logic for limiting `openTabs` relied on `reverse() + splice(10, 1000) + reverse()`.  
This worked only if the total number of open tabs never exceeded 1010.  
If a user accumulated more than 1000 tabs (e.g. 1140 over a year of usage), the result would incorrectly keep 140 tabs instead of just the last 10.

This PR replaces that logic with a simpler and correct approach:
```js
openTabs = openTabs.slice(-10);


hopleus:

PR #1026, opened 2025-10-07
[Bug, EC] PEES-771: Bug fix too much recursion when uploading lots of file

kingjia90:
Resolves https://github.com/pimcore/service-operations/issues/234

### DIFF
--- a/public/js/pimcore/asset/tree.js
+++ b/public/js/pimcore/asset/tree.js
@@ -363,7 +363,11 @@
  
          if(dataTransfer["items"] && dataTransfer.items[0] && dataTransfer.items[0].webkitGetAsEntry) {
              // chrome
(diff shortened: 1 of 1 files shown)


public/js/pimcore/asset/tree.js
```diff
@@ -363,7 +363,11 @@
  
          if(dataTransfer["items"] && dataTransfer.items[0] && dataTransfer.items[0].webkitGetAsEntry) {
              // chrome
-             var traverseFileTree = function (item, path) {
+             var traverseFileTree = function (item, path, depth = 0) {
+                 if (depth > 100) {
```
kingjia90: Not sure what value could make more sense or if it would crash as well before that, maybe
```suggestion
                 if (depth > 50) {
```

PR #1028, opened 2025-10-08
[Bug, EC] PEES-751: Fix Classification Store Groups pagination

kingjia90:
Resolves https://github.com/pimcore/service-operations/issues/247


kingjia90:
draft: ops, wrong spot to apply the fix

PR #1030, opened 2025-10-09
[Improvement]: Include Pimcore core version in JSON export filenames for better version traceability

kingjia90:
Resolves https://github.com/pimcore/internal-improvements/issues/6


kingjia90:
Actually need to adapt the Core as well
https://github.com/pimcore/pimcore/pull/18750

and probably all the other exportable stuff like Field collection, bricks etc..

PR #1034, opened 2025-10-14
[IMPROVEMENT] Classification store: add title to key/group search popup

dturopoli:
Added title column to key/group search

Screenshot_79


kingjia90:
The issues is that there would be no minor releases on admin ui due to its end of life in favor of Studio, so could maybe merged as a bugfix just for it's minimal change but needs to be rebased first

kingjia90:
Closing it, we won't accept improvement for already in end of life  admin ui classic anymore

PR #1041, opened 2025-10-23
[Bug, EC] PEES-818: Bug: Relation Filter Using Tags

kingjia90:
Resolves https://github.com/pimcore/service-operations/issues/272

### DIFF
--- a/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -422,14 +422,14 @@ pimcore.element.helpers.gridColumnConfig = {
 
             const activeFilter = this.grid.getStore().getFilters().items;
 
-            for (let i = 0; i < activeFilter.length; i++) {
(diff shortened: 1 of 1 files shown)


public/js/pimcore/element/helpers/gridColumnConfig.js
```diff
@@ -418,7 +418,21 @@ pimcore.element.helpers.gridColumnConfig = {
                 containerType: "filterByRelationWindow"
             });
 
-            editor.fieldConfig.width = 300;
```
kingjia90, line 421: unrelated change, but this 300 seems off 🤔

robertSt7: @kingjia90 I think we then re-introduce this bug, right? https://github.com/pimcore/admin-ui-classic-bundle/pull/888

kingjia90: ah.. reverting the change here as unrelated to the PR topic.
To me there's something odd concerning the popup  window auto-resize in general, but it's likely something under the hood of ExtJS

PR #1043, opened 2025-10-27
[Bug]: Fix quantity value IN search

kingjia90:
When using this filter
image 
on a quantity value (or RGB color dataype), it is false positively seen as array of ids entering in this block, due to the nature of `array(value,unit)` and breaking as it would try to compare with an `IN` operator on a single value.

This change generally makes also sure if there's just one element in the list to opt for equal operator instead of `IN`


kingjia90:
Tests failing due to Symfony 7 on Core 12.x

kingjia90:
Merging in the meanwhile, for tests would be fixed separately and since the other are just passing, no red flags so far

PR #1050, opened 2025-11-05
[Bug, EC] PEES-794: Enterprise Asset Metadata | Saved Grid Options only available in one folder

robertSt7:
resolves https://github.com/pimcore/service-operations/issues/250
Make saved Grid Options in assets available for all folder

also see https://github.com/pimcore/pimcore/pull/18817

### DIFF
--- a/src/Controller/Admin/Asset/AssetHelperController.php
+++ b/src/Controller/Admin/Asset/AssetHelperController.php
@@ -60,7 +60,7 @@ class AssetHelperController extends AdminAbstractController
         $db = Db::get();
         $configListingConditionParts = [];
         $configListingConditionParts[] = 'ownerId = ' . $userId;
-        $configListingConditionParts[] = 'classId = ' . $db->quote($classId);
+        $configListingConditionParts[] = '(shareBetweenFolders = 1 or classId = ' . $db->quote($classId) . ')';
 
         if ($searchType) {
             $configListingConditionParts[] = 'searchType = ' . $db->quote($searchType);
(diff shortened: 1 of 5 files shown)


src/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -245,8 +246,8 @@ public function doGetGridColumnConfig(Request $request, bool $isDelete = false):
             return ($a['position'] < $b['position']) ? -1 : 1;
         });
 
-        $availableConfigs = $classId ? $this->getMyOwnGridColumnConfigs($userId, $classId, $searchType) : [];
-        $sharedConfigs = $classId ? $this->getSharedGridColumnConfigs($this->getAdminUser(), $classId, $searchType) : [];
+        $availableConfigs = $classId ? $this->getMyOwnGridColumnConfigs($userId, $searchType) : [];
+        $sharedConfigs = $classId ? $this->getSharedGridColumnConfigs($this->getAdminUser(), $searchType) : [];
```
kingjia90:
```suggestion
        $availableConfigs = $this->getMyOwnGridColumnConfigs($userId, $searchType);
        $sharedConfigs = $this->getSharedGridColumnConfigs($this->getAdminUser(), $searchType);
```

kingjia90 (commented):
Overall LGTM, my only doubt is more at a bigger picture, should this new behavior be configurable in the grid save options? 

In other words, some config would be better fit to be limited to a given folder (intentional purpose), and for existing project might have suddenly too many grid configuration (and potentially with the same name, if they manually worked around the current/former limitation) after PR.

image

Something like `Share across all assets folder` and keep the current behavior still as default

Alternatively, it could also be something like  `Apply to subfolders` (a sort of inheritance) and add possibility to save grid options to to `root` and parent folders and we traverse the classIds/folders

Extra: check how Studio will tackle this

kingjia90 (commented):
In Listing, it might need to cast to boolean like in
https://github.com/pimcore/admin-ui-classic-bundle/blob/6041e7d2f89dee6d279b46fdd6b7d8d6c7784e7c/src/Model/GridConfig/Listing/Dao.php#L35
to be consistent returning `false` when `null`

The rest LGTM

src/Model/GridConfig.php
```diff
@@ -47,6 +47,8 @@ class GridConfig extends AbstractModel
 
     protected ?bool $saveFilters = null;
 
+    protected ?bool $shareBetweenFolders = null;
```
kingjia90, line 50: should we make the property and the setter/getter non-nullable like the others `shareGlobally` and `setAsFavourite`?

robertSt7: @kingjia90 In my opinion we should keep it nullable, because the column in the database is also nullable. What would be the benefit of making it different?

kingjia90: That would be also fine to me, was just for consistency.
IIRC we used to have this approach that `null` is undefined only on insert (before first save), but on read/update/unserialize or so, it would be `false`, but not sure anymore if it's still valid or that strict.

PR #1055, opened 2025-11-18
[Grid] Bugfix: Save relation filter in grid config

BlackbitDevs:
After #963, saving relation filters in a grid config does not work anymore. This PR fixes this (while keeping the enhancements of #963).

Furthermore, this PR adds the functionality that filtered relation columns also get displayed in italic (like "normal" filtered columns).

### DIFF
--- a/src/Controller/Admin/MiscController.php
+++ b/src/Controller/Admin/MiscController.php
@@ -88,7 +88,7 @@ class MiscController extends AdminAbstractController
             $fallbackLanguages[] = \Locale::getPrimaryLanguage($language);
         }
         if ($language != 'en') {
-            $fallbackLanguages[] = \Locale::getPrimaryLanguage(Tool::getDefaultLanguage());;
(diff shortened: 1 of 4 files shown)


robertSt7 (approved):

src/Controller/Admin/MiscController.php
```diff
@@ -88,6 +88,7 @@ public function jsonTranslationsSystemAction(Request $request, TranslatorInterfa
             $fallbackLanguages[] = \Locale::getPrimaryLanguage($language);
         }
         if ($language != 'en') {
+            $fallbackLanguages[] = \Locale::getPrimaryLanguage(Tool::getDefaultLanguage());;
```
robertSt7:
```suggestion
            $fallbackLanguages[] = \Locale::getPrimaryLanguage(Tool::getDefaultLanguage());
```

robertSt7:
@BlackbitDevs Thank you for creating this PR, but we can't merge this without an signed CLA. We will take care of this in studio

PR #1060, opened 2025-11-26
[Bug, EC] PEES-796: Fix relations dialog out of scope

kingjia90:
Resolves https://pimcore.atlassian.net/browse/PEES-796, https://github.com/pimcore/service-operations/issues/262

Kind of follow up https://github.com/pimcore/admin-ui-classic-bundle/pull/888#issuecomment-2909559203, that change in the PR was too general/broad and broke all the sizing. It worked just for many to one because it matches this fallback https://github.com/pimcore/admin-ui-classic-bundle/blob/1a6a1aff055f20e475189b98cbe5b180e93bf32c/public/js/pimcore/object/tags/manyToOneRelation.js#L160

Notes: not the cleanest code but it's for sunsetting admin ui and extJS seems really *ugh* to manage sizing, feels like fixing "bad" legacy code with slight less worse one 😅

### DIFF
--- a/public/js/pimcore/element/helpers/gridCellEditor.js
+++ b/public/js/pimcore/element/helpers/gridCellEditor.js
@@ -91,9 +91,20 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
             bodyStyle: "padding: 10px;"
         });
         let width = 700;
-        if (tagType === 'manyToManyObjectRelation' && fieldInfo.layout.width && fieldInfo.layout.width !== '100%') {
+
+        if (tagType === 'manyToOneRelation') {
+            width = sumWidths(
+                formPanel.config.items[0].items.items[0].width,
+                formPanel.config.items[0].labelWidth + 140
+            );
+        } else if (
+            tagType.endsWith('Relation') &&
+            fieldInfo.layout?.width &&
+            fieldInfo.layout?.width !== '100%'
+        ) {
             width = sumWidths(fieldInfo.layout.width, 25);
         }
+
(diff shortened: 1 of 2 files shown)


public/js/pimcore/element/helpers/gridCellEditor.js
```diff
@@ -91,9 +91,20 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
             bodyStyle: "padding: 10px;"
         });
         let width = 700;
-        if (tagType === 'manyToManyObjectRelation' && fieldInfo.layout.width && fieldInfo.layout.width !== '100%') {
+
+        if (tagType === 'manyToOneRelation') {
+            width = sumWidths(
+                formPanel.config.items[0].items.items[0].width,
```
kingjia90, line 97: this is like this because of a "hidden" default, so getting the `href.width` would make it match 1:1
https://github.com/pimcore/admin-ui-classic-bundle/blob/1a6a1aff055f20e475189b98cbe5b180e93bf32c/public/js/pimcore/object/tags/manyToOneRelation.js#L160

public/js/pimcore/element/helpers/gridCellEditor.js
```diff
@@ -91,9 +91,20 @@ Ext.define('pimcore.element.helpers.gridCellEditor', {
...
+                formPanel.config.items[0].labelWidth + 140
```
kingjia90, line 98: add label's size plus 3 buttons as 140 😓

PR #1069, opened 2025-12-14
Add Dedicated ManyToManyAssetRelation

cancan101:
Admin work for: https://github.com/pimcore/pimcore/pull/18879


kingjia90:
Thank you for your PR! Unfortunately, i must closing this one for the moment, it wouldn't be supported in admin ui classic because it's sun-setting and we won't tag any further minor releases. 
The PR in the Core could be good for Studio.

If you have any question or follow-up, please post/reply on PEES-897

PR #1070, opened 2025-12-16
[Bug, EC] PEES-888: Loss of all relation data in FC after moving DataObjects

kingjia90:
Resolves https://pimcore.atlassian.net/browse/PEES-888, https://github.com/pimcore/service-operations/issues/306

Field collection relations might be not fully initialized without forcing it, along with the "problem" that for the relations in a FC are not getting Delta calculated, but they get deleted and reinserted in the DB during update process.

### DIFF
--- a/src/Controller/Admin/DataObject/DataObjectController.php
+++ b/src/Controller/Admin/DataObject/DataObjectController.php
@@ -1043,14 +1043,14 @@ class DataObjectController extends ElementControllerBase implements KernelContro
         if (is_array($ids)) {
             $return = ['success' => true];
             foreach ($ids as $id) {
-                $object = DataObject::getById((int)$id);
+                $object = DataObject::getById((int)$id, ['force' => true]);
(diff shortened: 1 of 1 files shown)


kingjia90:
More details, the runtime cached object doesn't have relations https://github.com/pimcore/pimcore/blob/97b0a04c6ed7d9d945428b75012fc271a99a3a25/models/DataObject/AbstractObject.php#L224

because is registered before https://github.com/pimcore/pimcore/blob/97b0a04c6ed7d9d945428b75012fc271a99a3a25/models/DataObject/AbstractObject.php#L232-L235

Alternatively we need to add this getRawRelationData to the non-forced one before returning object on
https://github.com/pimcore/pimcore/blob/97b0a04c6ed7d9d945428b75012fc271a99a3a25/models/DataObject/AbstractObject.php#L204

kingjia90:
Closing in favor of https://github.com/pimcore/pimcore/pull/18892 for a more widespread coverage

PR #1084, opened 2026-01-26
Bug: Add html to display properties as part of the version data.

torqdev:
Properties are versioned data, and therefore if you publish a previous version it will revert back to a previous set of properties. Properties are not included in the version tool, and therefore make it seem like these are not part of the versioned data.


robertSt7 (approved):

robertSt7:
@torqdev Thanks for adding the properties to the version data

PR #1087, opened 2026-01-28
Fix SimpleBackendSearch Error 414 - Request-URI Too Large

t-engl:
Use POST instead of GET to search for objects for _combo_ to prevent 414 error response if possible.

The change should be backwards compatible, because the used Http-Method is declared by the actual implemenation and defaulting to GET. 

For this PR to take effect, SimpleBackendSearch bundled with pimcore 12.x itself should also be patched, but should not break even without the change.

## Additional info
It may be sufficient to remove the extra parameter _unsavedChanges_ instead, which is most likelly the root cause of this issue, but I don't know it's purpose.


t-engl:
Added related PR to [pimcore/pimcore](https://github.com/pimcore/pimcore/pull/18955)

PR #1110, opened 2026-05-13
[Bug]: Dataobject localized block items overridden randomly (off by one error?)

### LINKED ISSUE #1073: [Bug]: Fallback Language for inherited localized fields not working.
### Pimcore version

v2025.3 & older

### Steps to reproduce

Create Dataclass with localized field containing wysiwyg field, inheritance enabled.

Image

Configure Fallback languages, e.g. DE fallsback to EN:

Image

Create a folder and create nested dataobject instances inside. Set localized wysiwyg field on parent:

Image

Open Datagrid view on parent folder level, add localized wysiwyg field to grid view options and switch to locale DE:

Image

### Actual Behavior

Column `localized.wysiwyg` `DE` does not inherit `localized.wysiwyg` `EN` `fallbackLanguage` values correctly. Interesting is that in Grid view the lowest nested object seems to have it, but the parent doesn`t, see screenshots in `how to reproduce` section. Also currently configured on [demo.pimcore.com/admin ](https://demo.pimcore.com/admin/login/deeplink?object_1165_folder) 

When exporting `CSV` no columns inherit the fallbacklanguage values. Also interesting, for locale`FR` no fallback language has been configured via system settings, but it behaves exactly like `DE` which has `EN` set as fallback. 

### Expected Behavior

`localized` `inherited` fields receive correct `fallback language` values if available.

robertSt7:
resolves #1073

### DIFF
--- a/public/js/pimcore/object/tags/localizedfields.js
+++ b/public/js/pimcore/object/tags/localizedfields.js
@@ -542,6 +542,10 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
                 continue;
             }
 
+            if (!this.languageElements[currentLanguage]) {
+                continue;
+            }
+
             for (var s = 0; s < this.languageElements[currentLanguage].length; s++) {
                 try {
                     if (ignoreIsDirty || this.languageElements[currentLanguage][s].isDirty()) {
--- a/public/js/pimcore/object/tags/localizedfields.js
+++ b/public/js/pimcore/object/tags/localizedfields.js
@@ -529,10 +529,18 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
 
         for (var i = 0; i < this.frontendLanguages.length; i++) {
             currentLanguage = this.frontendLanguages[i];
-            if (!this.languageElements[currentLanguage]) {
+            localizedData[currentLanguage] = {};
+
+            // If the language tab was never rendered (deferred rendering, user never visited it),
+            // languageElements is empty. Preserve the original data to avoid wiping it when the
+            // whole block becomes dirty due to an unrelated add/remove of a sibling block item.
+            if (
+                    (!this.languageElements[currentLanguage] || this.languageElements[currentLanguage].length === 0) &&
+                    this.data[currentLanguage]
+               ) {
+                localizedData[currentLanguage] = Ext.clone(this.data[currentLanguage]);
                 continue;
             }
-            localizedData[currentLanguage] = {};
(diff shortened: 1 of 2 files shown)


public/js/pimcore/object/tags/localizedfields.js
```diff
@@ -547,6 +547,13 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
 
                 }
             }
+            
+            // If the language tab was never rendered (deferred rendering, user never visited it),
```
kingjia90: Just noticed that maybe we can do a little re-ordering here, so that the `for (var s)` is skipped with an early continue if length is 0

```
 for (var i = 0; i < this.frontendLanguages.length; i++) {
    currentLanguage = this.frontendLanguages[i];

    if (!this.languageElements[currentLanguage]) {
        continue;
    }

    localizedData[currentLanguage] = {};

    // Deferred-rendered tab: preserve original data
    // If the language tab was never rendered (deferred rendering, user never visited it),
    // languageElements is empty. Preserve the original data to avoid wiping it when the
    // whole block becomes dirty due to an unrelated add/remove of a sibling block item.
    if (
        this.languageElements[currentLanguage].length === 0
        && this.data[currentLanguage]
    ) {
        localizedData[currentLanguage] = Ext.clone(this.data[currentLanguage]);
        continue;
    }


    for (var s = 0; s < this.languageElements[currentLanguage].length; s++) {
        try {
            if (ignoreIsDirty || this.languageElements[currentLanguage][s].isDirty()) {
                localizedData[currentLanguage][this.languageElements[currentLanguage][s].getName()]
                    = this.languageElements[currentLanguage][s].getValue();
            }
        } catch (e) {
            console.log(e);

            localizedData[currentLanguage][this.languageElements[currentLanguage][s].getName()] = "";
        }
    }
}

```

kingjia90: My only "concern" is if `
    if (!this.languageElements[currentLanguage]) {
        continue;
    }`  should be also considered as part of deferred rendered silently failed case, and be falling back to `this.data[currentLanguage]`

In other words
```
if ((!languageElements || languageElements.length === 0) && this.data[currentLanguage]) {
    localizedData[currentLanguage] = Ext.clone(this.data[currentLanguage]);
    continue;
}

```

kingjia90 (changes requested):
Overall LGTM, i haven't reproduced it but looks the right follow up for https://github.com/pimcore/admin-ui-classic-bundle/pull/1059

PR #1133, opened 2026-07-14
[Bug, EC] PEES-1176: Fix dependencies tab pagination for restricted u…

jcPimcore:
Resolves https://github.com/pimcore/service-operations/issues/851

Depends on: https://github.com/pimcore/pimcore/pull/19257

Required By pagination filtered permissions after the SQL LIMIT/OFFSET, so a restricted user's page count matched an admin's while individual pages could come back sparse or empty, and the ExtJS grid treated a zero-row page as the end of the dataset, blocking navigation to later pages that had visible items.

- getRequiredByDependenciesAction now scans past hidden rows to fill each page, re-deriving the raw-to-visible mapping from scratch on every request so paging stays correct across independent calls, and reports an exact permission-filtered total via the newly cached pimcore/pimcore Service::getRequiredByVisibleTotalCount().
- getRequiresDependenciesAction filters the already fully-loaded in-memory requires set before slicing, giving an exact total at no extra cost.
- dependencies.js refreshes the hidden-dependencies note on every page load instead of only the first.
- overrides.js adds a narrow Ext.toolbar.Paging fix so navigation isn't disabled when a page is empty but the store's total indicates more pages exist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "endroid/qr-code": "^6.0.1",
         "nesbot/carbon": "^3.8.4",
         "phpoffice/phpspreadsheet": "^2.2 || ^3.3",
-        "pimcore/pimcore": "^12.3",
+        "pimcore/pimcore": "^12.3.12",
         "symfony/webpack-encore-bundle": "^1.17 || ^2.0"
     },
     "conflict": {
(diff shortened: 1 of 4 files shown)


jcPimcore:
The `static-analysis` CI failure here is expected and not a bug in this PR.

This branch calls two new `Pimcore\Model\Element\Service` methods (`getRequiredByVisibleTotalCount()` and `getRequiredByHasHiddenDependencies()`) added in pimcore/pimcore#19257. That core PR hasn't been merged/tagged yet, and this bundle's `composer.json` pins `pimcore/pimcore: ^12.3`, which resolves against Packagist to the latest tagged release (currently `v12.3.6`) rather than the `12.3` branch tip - so PHPStan correctly can't find methods that only exist as unmerged commits on that branch.

Once pimcore/pimcore#19257 is merged into `12.3` and a new tag is cut that includes it, re-running CI here should pass without any further changes to this PR.

